PR #31, opened 2017-02-24
Added lock button to switch between edit/read-only mode

### LINKED ISSUE #7: Add a lock snippet button that locks the snippet into place
This would stop the parsing and enter "Annotation Mode"

solkaz:
Fixes #7

### DIFF
--- a/src/components/AppBody.jsx
+++ b/src/components/AppBody.jsx
@@ -56,6 +56,10 @@ class AppBody extends React.Component {
   }
 
   switchReadOnlyMode() {
+    if (this.state.readOnly) {
+      console.log("Switching back to editing mode not supported yet");
+      return;
+    }
     this.setState({
       isDialogOpen: false,
       readOnly: !(this.state.readOnly),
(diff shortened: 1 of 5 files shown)


thebho:
Can you provide a screenshot?

src/components/AppBody.jsx
```diff
@@ -44,6 +47,12 @@ class AppBody extends React.Component {
     });
   }
 
+  switchReadOnlyMode() {
+    this.setState({
+      readOnly: !(this.state.readOnly),
```
dane-johnson: I think this should be one way. You lock it in, it's locked in for good.

src/components/LockButton.jsx
```diff
@@ -0,0 +1,20 @@
+import React, { PropTypes } from 'react';
+
+import FlatButton from 'material-ui/FlatButton';
+
+
+const LockButton = ({onClick, readOnly}) => {
+  return (
+    <FlatButton
+      label={`Switch to ${ readOnly ? 'edit' : 'read-only' } mode`}
```
dane-johnson: Which would make this disappear when locked in.

src/components/AppBody.jsx
thebho: Right, there is no "edit snippet" path after locking in this iteration

PR #40, opened 2017-02-24
Save snippet button

### LINKED ISSUE #10: Add a "Save Snippet" button that allows saving the state to Redis
Save Snippet will save the entire state of the snippet.
1. Open a modal that allows you to name and confirm the snippet
2. Confirm posts to redis that

jdmaryville:
Add save snippet button which saves snippet state to redis.

fixes #10


src/components/TokenSelector.jsx
```diff
@@ -6,27 +6,35 @@ import Checkbox from 'material-ui/Checkbox';
 class TokenSelector extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {
```
solkaz: You should be using `this.state` and `this.setState` to manage the state of the component

solkaz: Really, this component could easily be a functional component. It doesn't need local state, and other components will need its state (for highlighting certain tokens, etc). I would move the state up to `AppBody` and pass down the state as a prop

src/components/SnippetArea.jsx
```diff
@@ -18,13 +20,18 @@ const snippetEditorModes = {
   python3: 'python',
 };
 
-const SnippetArea = ({ contents, isDialogOpen, onTitleChanged, onSnippetChanged, readOnly, switchReadOnlyMode, snippetLanguage, toggleConfirmLockDialogVisibility }) => {
+
+const SnippetArea = ({ contents, isDialogOpen, onTitleChanged, onSnippetChanged,
```
solkaz: The amount of props we're passing down is getting ridiculous (not wrong, just annoying).

src/components/TokenSelector.jsx
```diff
@@ -6,27 +6,35 @@ import Checkbox from 'material-ui/Checkbox';
...
       return (
         <ListItem
-          key={`${tokenType.text}-index`}
-          leftCheckbox={<Checkbox />}
-          primaryText={tokenType.text}
+          key={`${tokenText}-index`}
+          leftCheckbox={
+            <Checkbox onCheck={
+                (ev, checked) => this.itemState[tokenText] = checked
+              }
```
solkaz: This should be a call to `setState`. You never want to directly modify the state object

src/components/TokenSelector.jsx
jdmaryville: Please approve

PR #42, opened 2017-02-27
Added Live AST Highlighting

### LINKED ISSUE #26: Add AST Chunker
The code for obtaining an AST of live input now exists. The JSON chunking code from the `demo_ui` project needs to be ported in order to process those ASTs.

Hopding:
Fixes #26. This PR adds live highlighting as code is entered. Note that token colors are hard coded, and there are a number of tokens that _could_ exist in code that have not been assigned colors in this PR. In order to support them, it is only necessary to add a line to the `tokenColors` map in the `src/util/highlight.js` file. I think adding support for additional tokens is a significant task, and should be the focus of a different story.

screen shot 2017-02-27 at 7 52 39 am


src/components/AppBody.jsx
```diff
@@ -73,10 +87,8 @@ class AppBody extends React.Component {
   }
 
   // Callback to be invoked when user edits the code snippet
-  onSnippetChanged(snippetContents) {
-    this.setState({
-      snippetContents
-    })
+  onSnippetChanged(snippet, codeMirrorRef) {
+    this.setState({ snippet });
```
solkaz: This is erroneous, it should be
```
this.setState({
  codesplainObj: {
    snippet,
  },
});
```

src/components/SnippetArea.jsx
```diff
@@ -2,54 +2,61 @@ import React, { PropTypes } from 'react';
...
+        <ConfirmLockDialog
+          isOpen={this.props.isDialogOpen}
+          accept={this.props.switchReadOnlyMode}
+          reject={this.props.toggleConfirmLockDialogVisibility}
+        />
+        <CodeMirror
+          ref={cm => this.codeMirror = cm}
+          value={this.props.contents}
+          options={this.codeMirrorOptions}
+          onChange={ev => this.props.onSnippetChanged(ev, this.codeMirror)}
```
solkaz: I would prefer to move this to its own function on the class, so that we're not creating a new function each time we render.

dane-johnson:
Need
- [ ] Function Def
- [ ] Class Def
- [ ] Try-catch
- [ ] Needs to be faster, too laggy

src/components/SnippetArea.jsx
Hopding: Not too sure what you're talking about here.

src/components/SnippetArea.jsx
```diff
@@ -2,54 +2,61 @@ import React, { PropTypes } from 'react';
...
-      />
-      <CodeMirror
-        value={contents}
-        options={codeMirrorOptions}
-        onChange={onSnippetChanged}
-      />
-      </CardText>
-    </Card>
-  );
+class SnippetArea extends React.Component {
```
Hopding: I thought you needed to have a class component to use refs. Turns out you don't. I've changed it back to a functional component.

thebho:
Dane, those seem like different stories/features to iterate over.  The requirement is live syntax rendering.

dane-johnson:
Right now it's not really live once you get down to about 14 lines.

thebho:
Is that a bug or just room for improvement?

PR #49, opened 2017-02-27
Devstack fix

dane-johnson:
Fixes bug in devstack that would cause it to fail if the redis container already exists

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+# redis
+/data/redis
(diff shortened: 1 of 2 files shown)


dane-johnson:
Wouldn't this fail if it existed but wasn't running?

thebho:
It just logs the error and the variable stays false

dane-johnson:
@thebho Not gonna lie, having a pretty hard time with this.

thebho:
It's a little different because you aren't using a compose file.  We can chat about it tomorrow at MT

dane-johnson:
Can now confirm this does indeed work

PR #61, opened 2017-02-28
Shrinks the TokenSelector component

### LINKED ISSUE #39: Shrink the TokenSelector component
It has too much real estate. Shrink it down/move it/whatever

jdmaryville:
Decreased size of token selector panel, increased the size of snippet area.

fixes #39

### DIFF
--- a/src/components/AppBody.jsx
+++ b/src/components/AppBody.jsx
@@ -155,7 +155,7 @@ class AppBody extends React.Component {
     return (
       <div className="container-fluid">
         <div className="row">
-          <div className="col-md-3">
+          <div className="col-md-2">
             <Card>
               <CardText>
                 <LanguageSelector
@@ -170,7 +170,7 @@ class AppBody extends React.Component {
               </CardText>
             </Card>
           </div>
-          <div className="col-md-5">
+          <div className="col-md-6">
             <SnippetArea
               title={this.state.snippetTitle}
               onSaveClick={this.onSaveState}
--- a/src/components/AppBody.jsx
+++ b/src/components/AppBody.jsx
@@ -1,13 +1,10 @@
 import React from 'react';
 import { browserHistory } from 'react-router'
-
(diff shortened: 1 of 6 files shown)


src/components/AppBody.jsx
```diff
@@ -140,6 +140,7 @@ class AppBody extends React.Component {
       setTimeout(() => parseReady = true, 1000);
       new Promise((resolve) => resolve(parser(snippet)))
       .then((AST) => {
+        console.log(JSON.stringify(AST));
```
Hopding: Should be able to take this guy out.

src/components/LanguageSelector.jsx
```diff
@@ -4,6 +4,12 @@ import MenuItem from 'material-ui/MenuItem';
 import SelectField from 'material-ui/SelectField';
 
 const LanguageSelector = ({ languages, onChange, selected }) => {
+  const styles = {
```
Hopding: I think we should move this, and the other inline styling, into a css file in the `/styles` folder.

src/components/AppBody.jsx
dane-johnson: I think I removed an identical line from a different PR. Make sure you guys are creating your branches off master.

dane-johnson:
http://www.material-ui.com/#/customization/styles

src/components/LanguageSelector.jsx
jdmaryville: I agree with moving the in-line styling into the css folder. But on the issue of width between tokens, wont there leave a pretty big amount of empty space between the last token and the length of the panel? Once we get more tokens, I believe then would be a better time to decrease the width between tokens.

src/components/AppBody.jsx
solkaz: We need to be more cognizant of taking out our `console.log` statements in our commits.

thebho:
Closing

PR #64, opened 2017-03-01
Add annotation creator and display components

### LINKED ISSUE #8: When selecting a "+" button for a code line open the annotation editor
The annotation panel/window would be editable. 

- [ ] Add a Save button that saves that annotation to the state "Annotations" array
- [ ] Click cancel to exit annotation.
- [ ] Clicking outside the box does nothing

### LINKED ISSUE #43: Make an annotation editor component

solkaz:
Fixes #8 
Fixes #43 

Adds annotation creator and display components


dane-johnson:
Another thing I noticed is that the text area is one line by default now, which is not immediately recognizable as and IDE to me.
screen shot 2017-03-01 at 10 58 53 am

PR #67, opened 2017-03-01
Fixed editor height

dane-johnson:
## Before
screen shot 2017-03-01 at 10 58 53 am
## After
screen shot 2017-03-01 at 11 18 21 am

PR #71, opened 2017-03-01
Editor and snippet line display instances of CodeMirror have their own respective heights

solkaz:
Each of these components have their respective heights; editor will be expanded, whereas the snippet line displays are compact and resize to fit their contents


dane-johnson:
Interesting that snapshot testing missed this. Maybe we need to re-evaluate our testing strategy

dane-johnson:
I believe this also closes #67

PR #77, opened 2017-03-02
Moved highlighting and parser logic from <AppBody /> to <SnippetArea />

### LINKED ISSUE #66: Syntax highlighting does not occur when returning to a page.
If you create a snippet and save, then return to that page, the syntax highlighting will be gone.

Links to #85

Hopding:
Fixes #66. The parser now runs every second, independently of changes to the snippet. I've also added a `triggerHighlight` method to `<SnippetArea />` that can be used to trigger a highlight of the snippet independently of the parser 'daemon' (note that it isn't actually a second thread, as JavaScript does not support true multithreading).


Hopding:
An update regarding concurrency in JavaScript: it looks like HTML5 does support truly concurrent processes in the form of "web workers": http://mobile.htmlgoodies.com/html5/tutorials/introducing-html-5-web-workers-bringing-multi-threading-to-javascript.html

(Note that these differ from Promises and Asynchronous functions, as asynchronicity doesn't necessarily imply a separate thread).

Will update this PR to use those web workers today...

thebho:
Does it stop that 1 second refresh after the snippet is "locked"

dane-johnson:
It does not, because otherwise I do not believe it would run when returning to a page with a saved snippet.

Hopding:
It continues to run the parser daemon (not yet really a daemon) every second, even if the snippet is locked. However, it won't actually re-parse the AST, because if the snippet is locked, then it cannot change. And ASTs are only generated if the snippet has changed.

dane-johnson:
👏

thebho:
Cool, will merge when conflicts are resolved

PR #80, opened 2017-03-02
Removed Golang parser

### LINKED ISSUE #60: Disable inactive parsers
I like that you can see the "Go" language selector, but we should disable that selection until we have that functionality

solkaz:
Fixes #60

### DIFF
--- a/src/components/AppBody.jsx
+++ b/src/components/AppBody.jsx
@@ -14,7 +14,6 @@ import { getTokenCount, getPrettyTokenName } from '../util/tokens.js';
 import axios from 'axios'
 
 const languages = [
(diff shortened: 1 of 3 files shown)


solkaz:
Do we want to also disable the LanguageSelector? Since they can only select one language

dane-johnson:
I say leave it. If nothing else it let's the user know that this is an ongoing project and there will be more languages in the future.

thebho:
Let's disable it.  And I'm wondering if it should be Java instead of Go, because that is next on the list

solkaz:
@thebho `LanguageSelector` is now disabled

PR #101, opened 2017-03-02
AnnotationCreator closes when saving

### LINKED ISSUE #97: After saving an annotation, the save and cancel buttons should go away and become close.
This is a regression.

solkaz:
Fixes #97

### DIFF
--- a/src/components/AnnotationCreator.jsx
+++ b/src/components/AnnotationCreator.jsx
@@ -29,6 +29,7 @@ class AnnotationCreator extends React.Component {
 
   saveAnnotation() {
     const { annotation } = this.state;
+    this.clearAnnotation();
     this.props.saveAnnotation(String(this.props.lineNumber), annotation);
   }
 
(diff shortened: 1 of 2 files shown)


thebho:
Does this close the whole panel, or just replace the Save/Cancel buttons?  The story is to remove the buttons

solkaz:
I thought I simply had to reimplement the behavior from before, so it closes the whole panel

solkaz:
I'm about to eat lunch but afterwards I should have this fixed

dane-johnson:
I think it should change to the display view

dane-johnson:
So basically
screen shot 2017-03-02 at 11 47 14 am
Then on save
screen shot 2017-03-02 at 11 47 25 am

solkaz:
That's the point of the story. I thought we were just reverting to previous behaior

thebho:
Staying in Edit Mode is the regression.  I was assuming in the previous story that "saving" would leave the annotation open but convert to View Mode.

solkaz:
@thebho @dane-johnson fixed

PR #121, opened 2017-03-02
SnippetArea contents state managed by AppBody

### LINKED ISSUE #118: Remove Snippet from State of SnippetArea Component
Use `this.props.content` within the parser daemon instead of `this.state.snippet`.

solkaz:
Fixes #118

### DIFF
--- a/src/components/SnippetArea.jsx
+++ b/src/components/SnippetArea.jsx
@@ -150,7 +150,7 @@ class SnippetArea extends React.Component {
         .forEach(t => {
           let selected = false;
           if (filters[t]) {
-            let selected = false;
+            selected = false;
           }
           newFilters[t] = {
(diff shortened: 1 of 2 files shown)


Hopding:
@thebho I think it would be good to merge this now. I doubt it is going to break anything, and it is fixing a significant error. Granted, you won't see any difference from a UX perspective.

solkaz:
I'll argue that this needs to be merged because we are calling `this.setState` in `<SnippetArea />`'s render method, which is [an incredibly bad idea](https://github.com/facebook/react/issues/5591), and something that should *never* happen. Moreover, we have the state of the CodeMirror's contents in two different places (in the local state _and_ in props); again, this is a bad idea, and goes against React's philosophy of having a single source of truth. Master may work "as-is", but merging this will yield performance benefits and maintain reliability of our code.

PR #129, opened 2017-03-03
Emphasize When Annotating a Line

### LINKED ISSUE #9: When looking at an annotation of, grey out the other other lines code to emphasize
Bold and Outlined

Hopding:
Fixes #9. This PR decreases the opacity of background text when annotating a line. I've also done some refactoring of the highlighting code.

## Not Making/Viewing Annotation
screen shot 2017-03-03 at 11 42 42 am

## While Making/Viewing Annotation
screen shot 2017-03-03 at 11 42 27 am


src/components/SnippetArea.jsx
```diff
@@ -131,54 +131,74 @@ class SnippetArea extends React.Component {
...
-          }
-        });
-
-      // Highlight the code snippet and invoke prop callback
-      highlight(snippet, AST, this.codeMirror, newFilters);
-      this.setState({
-        prevSnippet: snippet
-      });
-      this.props.onParserRun(AST, newFilters);
+      if (snippet !== undefined &&
```
solkaz: This can be condensed down. Both `''` and `undefined` are falsey in Javascript, so you could just use
```
if (!snippet || snippet === this.state.prevSnippet) {
   return;
}
```

solkaz: Which is nice because it's more compact, succinct, and needs to evaluate less things

src/components/SnippetArea.jsx
```diff
@@ -131,54 +131,74 @@ class SnippetArea extends React.Component {
...
+        const AST = parser(snippet);
+
+        // Get an array mapping token types to their occurence count
+        const tokenCount = [];
+        getTokenCount(AST, tokenCount);
+
+        // Generate array of strings containing pretty token name and its count
+        const filters = this.props.filters;
+        let newFilters = {};
+        Object.keys(tokenCount).filter(t => getPrettyTokenName(t) !== undefined)
```
solkaz: Not a fan of this indentation style

src/util/codemirror-utils.js
```diff
@@ -0,0 +1,135 @@
...
+    styleRegion(
+      codeMirror,
+      node.begin,
+      node.end,
+      `background-color: ${color};`
+    );
+  }
+
+  // Highlight all children of this token
+  node.children.forEach(child => {
```
solkaz: Avoiding recursion is going to help us with performance. If we were to use something like lodash's [_.forIn](https://lodash.com/docs/4.17.4#forIn), that would simplify the logic of what we're doing, and avoid recursion

src/components/SnippetArea.jsx
Hopding: Fixed .

src/util/codemirror-utils.js
Hopding: I think this is deserving of a different story, as this story isn't really about the performance of the parser/highlighting algorithms (though, I agree there is room for improvement in those areas).

thebho:
Is this still ready?

solkaz:
Yes

Issue #63, opened 2017-03-01
Update Python Parser


Hopding:
Update the parser packaged with the UI to include the changes introduced here: https://github.com/maryvilledev/codesplain/commit/32a38fa09bd93da0cba35aa57a1d7aab210b0b30

PR #157, opened 2017-03-07
Token documentation script 21

Hopding:
Fixes https://github.com/maryvilledev/codesplain/issues/21. Fixes https://github.com/maryvilledev/codesplainUI/issues/63. Run `npm run-script make-rule-csv` to generate the `rule-csv/python-rules.csv` file. This file contains information about all the rules recognized by the parser, and what happens to them at various stages in the parser and UI.

## How it looks when opened in Excel
screen shot 2017-03-07 at 4 30 48 pm


dane-johnson:
Add this step to Circle

Issue #38, opened 2017-02-24
Create Enzyme tests for components to ensure desired behavior


dane-johnson:

solkaz:
https://github.com/Syncano/syncano-dashboard

PR #168, opened 2017-03-10
Added Redux

dane-johnson:
Top level state is now managed by Redux. Components in charge of state management and async calls are now referred to as 'Containers'.


solkaz:
Fixes #38
Fixes #15

Issue #103, opened 2017-03-02
Disable the animation on language selector until we actually have another language to select


thebho:
Clicking on the triangle should do nothing.

solkaz:
There isn't a prop to set that in Material-UI. However, looking into the `<SelectField />` component, it seems that it's just a wrapper around some other components (such as MenuItem and IconButton), so either we could hand-roll our own Select component, or possibly make our own fork of Material-UI to make modifications to

solkaz:
I made an issue for the repo, and apparently it is a bug; I'm currently working on a fix for it in the Material-UI repo. So this isn't something that's dependent on our code, it's in Material-UI

solkaz:
Once [this PR](https://github.com/callemall/material-ui/pull/6286) is merged, we can close this issue

solkaz:
PR has been merged; closing this

PR #169, opened 2017-03-13
Updated packages

solkaz:
Updates Material-UI to 0.17.1, which includes my PR to fix [this issue](https://github.com/maryvilledev/codesplainUI/issues/103)

Also updates other packages that we use


solkaz:
React-router just did a major version upgrade; removing it

PR #176, opened 2017-03-13
Highlight moved to componentDidUpdate

### LINKED ISSUE #174: Restore Snippet Highlighting When Filters Are  Changed
A regression has occurred during the redux branch merge, which removed the immediate triggering of a snippet re-highlight when the state of the filters changes. The result is a noticeable delay in the re-highlighting of a snippet in response to a filter toggle, because the highlighting is only performed by the parser "daemon" every second.

A potential fix for this would be to subscribe to changes in the state of the filters in the redux store, and trigger a re-highlight when it is updated.

### LINKED ISSUE #184: Restore Emphasization When Annotating/Viewing a Line
A regression has occurred during the redux branch merge which removes the changes put into effect from https://github.com/maryvilledev/codesplainUI/pull/129.

dane-johnson:
Closes #174 
Closes #184

### DIFF
--- a/src/components/Editor.jsx
+++ b/src/components/Editor.jsx
@@ -61,6 +61,9 @@ class Editor extends React.Component {
     const {
       markedLines,
       openLine,
+      value,
+      AST,
+      filters
     } = this.props;
 
     const codeMirrorInst = this.codeMirror.getCodeMirror();
@@ -72,6 +75,13 @@ class Editor extends React.Component {
     this.deEmphasize();
     if (openLine) {
       this.emphasizeLine(openLine);
+    }
+    if (value && AST.children && filters) {
+      try {
+        highlight(codeMirrorInst, AST, filters)
+      } catch (err) {
+        console.error(err)
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/Editor.jsx
```diff
@@ -73,6 +76,13 @@ class Editor extends React.Component {
     if (openLine) {
       this.emphasizeLine(openLine);
     }
+    if (value && AST.children && filters) {
```
Hopding: This looks like it will override the annotation highlighting that is done by `this.emphasizeLine`. 

I just tested annotation highlighting in this branch and master, and when a line is being annotated/viewed, it is no longer emphasized (another regression). So it won't be possible to know for sure if this will interfere with it or not, as it doesn't actually work anymore. 

However, I think adding a condition that verifies no lines are currently being emphasized will prevent this fix from interfering with the `emphasizeLine` regression fix when it is put in place.

dane-johnson: That's odd, I seem to remember it working fine before the merge. Will look into that.

PR #182, opened 2017-03-13
Fixed by increasing bootstrap class and making text checkbox label

### LINKED ISSUE #135: Fix the small window size spacing between token filters and their check boxes
Shrinking the window size should not cause the token name text to smash into and over the check box

dane-johnson:
Closes #135


dane-johnson:
screen shot 2017-03-13 at 4 20 47 pm

PR #183, opened 2017-03-13
Added "Save" & "Save As" Options

### LINKED ISSUE #73: Saving on a snippet that already has an ID should prompt to save or save as.
Add to toolbar

Hopding:
Fixes #73. This PR converts the `<SaveButton />` component to a `<SaveMenu />` component with "Save" and "Save As" options. 

-----

When the "Save" option is clicked:

1. If no ID exists in the URL (e.g. `localhost:8080`), then the snippet is saved under a new UUID.
2. If an ID does exist in the URL (e.g. `localhost:8080/8d12f0ca-ea50-4f15-ba73-65f0e8dddb52`), then the snippet is saved under that same ID.

When the "Save As" option is clicked:

1. The snippet is saved under a new ID.

-----

Note the URL changes in the following screenshots:
## Before Saving
screen shot 2017-03-13 at 5 27 32 pm

## Immediately After "Save"
screen shot 2017-03-13 at 5 27 38 pm

## Immediately After "Save As"
screen shot 2017-03-13 at 5 27 45 pm


dane-johnson:
I feel like this needs to await Github Auth. Otherwise, anyone can change any snippet provided they know the URL

Hopding:
We have GitHub Auth working, but it is not being used for anything. How would you like to see that tied in with this?

PR #192, opened 2017-03-14
Saves will no longer occur without a snippet title

### LINKED ISSUE #74: User should be prompted to enter a title for a snippet
When a user clicks the save button, if the snippet doesn't have a title, he/she should be prompted to enter a title

Hopding:
Fixes #74.

## Clicking Save Without Title
screen shot 2017-03-14 at 2 14 31 pm

## Clicking Save With Title
screen shot 2017-03-14 at 2 15 10 pm

### DIFF
--- a/src/containers/SnippetArea.jsx
+++ b/src/containers/SnippetArea.jsx
@@ -124,7 +124,7 @@ export class SnippetArea extends React.Component {
     const { snippetTitle, appState } = this.props;
     if (!snippetTitle) {
       this.setState({ titleErrorText: 'This field is required' });
-      this.showSnackbar('Please populate all required fields before saving.');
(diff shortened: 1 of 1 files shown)


src/containers/SnippetArea.jsx
```diff
@@ -114,7 +121,13 @@ export class SnippetArea extends React.Component {
   }
 
   handleSaveAs() {
-    const { appState } = this.props;
+    const { snippetTitle, appState } = this.props;
+    if (!snippetTitle) {
+      this.setState({ titleErrorText: 'This field is required' });
+      this.showSnackbar('Please populate all required fields before saving.');
```
dane-johnson: Since we know which field it is, why not specifically reference the title field in the snackbar error?

PR #196, opened 2017-03-15
Fixed speed issues by using deep comparison of props to prevent unnecessary updates

dane-johnson:
Lodash provides a deep comparison function that is fast yet robust enough to prevent unnecessary re-renders, significantly improving speed.


src/components/Editor.jsx
```diff
@@ -47,6 +48,10 @@ class Editor extends React.Component {
     this.deEmphasize = this.deEmphasize.bind(this);
   }
 
+  shouldComponentUpdate(nextProps) {
+    return _.isEqual(nextProps, this.props)
```
solkaz: https://facebook.github.io/react/docs/react-component.html#shouldcomponentupdate

What you're saying then is that the Editor component should not call `render()` unless the current and the next props are equal to each other. That doesn't seem correct to me. The child CodeMirror instance will still re-render because its state has changed (i.e. its value), but only re-rendering when the props have not changed seems counter-intuitive, and contrary to what React does.

solkaz: I do think we can utilize the `shouldComponentUpdate()` method to limit the number of re-renders, but compare specific props instead of the entire props object.

dane-johnson:
Oops, had that backwards. Meant to update only if properties **had** changed.

src/components/Editor.jsx
```diff
@@ -111,9 +115,9 @@ class Editor extends React.Component {
 
   render() {
     const {
-      onChange,
       readOnly,
       value,
+      onChange
```
solkaz: Please try to keep the props in order, and with trailing commas

dane-johnson: Order?

dane-johnson:
- [ ] Fix filter change regression

src/components/Editor.jsx
solkaz: Alphabetic order, that is.

src/util/codemirror-utils.js
```diff
@@ -41,11 +34,13 @@ function highlightNode(codeMirror, node, filters, parentColor) {
 
   // Highlight all children of this token
   node.children.forEach(child => {
-    if (child === Object(child)) 
+    if (child === Object(child))
       highlightNode(codeMirror, child, filters, color);
   });
 }
 
+const throttledHighlightNode = _.throttle(highlightNode, 500)
```
solkaz: Would debounce make more sense?
- https://lodash.com/docs/4.17.4#debounce
- https://css-tricks.com/debouncing-throttling-explained-examples/

dane-johnson:
I think the highlighter is still in need of additional optimizations, but I will open another story for that.

Hopding:
I ran this locally, and after reaching about 15 lines of code, it still becomes unusable slow. I'm not sure that this fixes the problem? Has anybody else tested this branch?

dane-johnson:
I think there is still an issue with panning that might have just emerged now. Looking in to that. Also if you run it with the highlighter off it is significantly faster. We may need to rethink the way we do the highlighting.

Hopding:
1. What do you mean by "panning"?
2. What do you mean by "with the highlighter off"?

When I tested it locally, I did not have any of the filters toggled, if that is what you mean. If you mean commenting the highlight function out so the function is not being called at all, I thought you had tested that before and did not notice a difference?

dane-johnson:
1. Panning, by which I mean when you reach the edge of the editor and the editor begins to scroll with the cursor into a new region.
2. I did mean with no filters on.

Hopding:
Just tested it again after commenting out lines 70, 74, and 78 of `Editor.jsx` (the calls to the emphasization and highlighting functions). Doing so seems to "fix" the problem entirely, and even after reaching 30 lines of code there is no performance issues. However, the same is true of the master branch.

So I would say it is clear that the problem has to do with the highlighting functions, or how (or the rate at which) we are using them. I do not mean to insult your work on this branch, but I do not believe that it fixes the problem (yet). 

I do not think we should optimize just for that sake of optimization. Unless the optimization results in an improved UX, it is probably not really necessary. Of course, it is clear that the highlighting stuff needs optimization, but we don't seem to yet know how exactly that ought to be done. But I do not think that anything else does.

Hopding:
@dane-johnson @thebho @solkaz @jdmaryville @JWalker1995 If anybody could confirm/deny my results, that would be useful. Just to be sure I'm not missing something.

dane-johnson:
You may be a few commits behind. The line numbers you listed don't correspond with the functions you describe.

Hopding:
My mistake. Those are the numbers for `master`. For the `speedFix` branch, they correspond to 75, 79, and 83.

src/components/Editor.jsx
```diff
@@ -78,14 +83,19 @@ class Editor extends React.Component {
       // would evaluate to false
       this.emphasizeLine(openLine);
     }
-    if (value && AST.children && filters) {
+    if (newAST || newFilters) {
```
solkaz: You need to flip these after the highlight occurs, or else it will keep highlighting it.

Hopding:
I just tested this again. I still do not notice a performance increase. And unless the code that was added in this branch actually improves the performance, I do not see a reason to merge it. I won't reject it though. I'd like to see if anybody else thinks the performance is improved by this branch, because my perception could be off. (or if we have a way to get hard numbers, that would be even better)

PR #198, opened 2017-03-15
Save to redis when annotation is saved

### LINKED ISSUE #102: When saving annotation on a "saved" code snippet, that should also update the snippet in Redis automatically

solkaz:
Fixes #102


src/actions/app.js
```diff
@@ -28,10 +30,31 @@ export const toggleEditState = () => ({
...
+    if (!id) {
+      // Return a Promise object to make this action "thenable"
+      return Promise.resolve();
+    }
+    // Get the app state and stringify it
+    const stateString = JSON.stringify(getState().app);
+
+    // TODO: Dispatch action to show the snackbar with a relevant message based on the results
+    // Return the results of the POST request
+    return axios.post(`api/snippets/${id}`, { json: stateString, })
```
dane-johnson: Shouldn't the container handle this? Why burden Redux with anything other than managing the state?

solkaz: [Because thunks](http://redux.js.org/docs/advanced/AsyncActions.html)

dane-johnson: Okay sure, but now there's coupling between updating state and backend interaction. The http requests are already async thanks to Promises, why not let the thing that's handling the event fire it off?

dane-johnson: There is coupling because that means every time an action is fired to save an annotation, a http request is created.

solkaz: Except there isn't, because the request will never affect the state being updated because it's just a _side effect_. If I had updated the state only if the request was successful, then yes, that is coupling, and the request is no longer a side effect, but that isn't the case.

dane-johnson: Why write create a Redux action with side effects? The container is meant to conduct data, it is a much more appropriate place to make an async data request.

Hopding:
I'd like to have an in-person discussion about this, because I'm not familiar enough with redux to know if this is good practice or not. But, it does seem to me that since we are managing all of the other networking via callbacks in container components, that it would make sense to continue with that approach (unless doing so is burdensome or impossible in this situation). I think doing so makes the code more consistent.

src/actions/app.js
solkaz: > Also, we need the state after the SAVE_ANNOTATION action. In the Annotations container, we only have the state before that action. We could merge the two together in the handler function, but that's the job of the reducer, not the container. By using thunk, we're able to call getState() after we've dispatched the SAVE_ANNOTATION action, so that we can easily save the state to Redis.

solkaz:
Why? It still is concerned with the App's state, thus why it is part of the App actions. I agree that we can make a story to transition all of the network requests to this model, but that's another story, and not part of the requirements for this story.

solkaz:
https://gist.github.com/markerikson/bb7589d395db3fe2a7bde5996ee7f193

solkaz:
http://andrewhfarmer.com/react-ajax-best-practices/

solkaz:
http://redux.js.org/docs/advanced/AsyncActions.html

thebho:
This works, anyone have any last argument why it shouldn't go through?

solkaz:
I think we're good. I'll let you merge this though.

Issue #175, opened 2017-03-13
Create an AWS API Gateway that can CRUD Snippets


thebho:
- [ ] Require API Key for dev environment
- [ ] Only allow access to dev bucket
- [ ] Path should be /{github_id}/{snippet_id}

Figure out how to route users a specific environment bucket

PR #203, opened 2017-03-16
add lambdas

thebho:
Connects #175

### DIFF
--- a/awslambda/deletesnippet/index.js
+++ b/awslambda/deletesnippet/index.js
@@ -1,0 +1,40 @@
+'use strict';
+
+console.log('Loading function');
+
+const aws = require('aws-sdk');
+
+const s3 = new aws.S3({ apiVersion: '2006-03-01' });
+
+
+exports.handler = (event, context, callback) => {
+    console.log(JSON.stringify(event));
+    var user_id = event.pathParameters.user_id;
+    var snippetKey = event.pathParameters.snippet_id;
+
+    var bucket;
+
+    // TODO: Add Other environments to process.env and if statement
+    if (event.requestContext.apiId === process.env.devID) {
+        bucket = process.env.devBucket;
+    } else {
+        console.err("Invalid apiId" + event.requestContext.apiId);
+    }
+
+    const params = {
+        Bucket: bucket,
+        Key: user_id + "/" + snippetKey,
+    };
+    s3.deleteObject(params, (err, data) => {
+        if (err) {
+            console.log(err);
+            const message = `Error deleteing object ${params.Key} from bucket ${bucket}.`;
+            console.log(message);
+            callback(message);
+            context.Error({ statusCode: 200});
+
+        } else {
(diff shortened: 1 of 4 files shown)


awslambda/deletesnippet/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+const aws = require('aws-sdk');
+
+const s3 = new aws.S3({ apiVersion: '2006-03-01' });
+
+
+exports.handler = (event, context, callback) => {
+    console.log(JSON.stringify(event));
+    var user_id = event.pathParameters.user_id;
+    var snippetKey = event.pathParameters.snippet_id;
```
dane-johnson: Use `const` in place of `var`. Also, 12 and 13 can be written as
`const {user_id, snippet_id as snippetKey} = event.pathParameters`

awslambda/deletesnippet/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+    var bucket;
```
dane-johnson: use `let` in place of `const`

awslambda/deletesnippet/index.js
dane-johnson: Additionally, I wouldn't mix camelCase and underscore naming

awslambda/getsnippet/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+
+exports.handler = (event, context, callback) => {
+    console.log(JSON.stringify(event));
+    var user_id = event.pathParameters.user_id;
+    var body = JSON.parse(event.body);
+    var snippetKey = body.snippetTitle;
+    console.log(snippetKey);
+
+    snippetKey = snippetKey.replace(/\s+/g, '_').toLowerCase();
```
dane-johnson: Is this to sanitize the input? Should probably be its own function, or at least be commented.

awslambda/deletesnippet/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+console.log('Loading function');
+
+const aws = require('aws-sdk');
+
+const s3 = new aws.S3({ apiVersion: '2006-03-01' });
+
+
+exports.handler = (event, context, callback) => {
+    let userId = event.pathParameters.user_id;
```
dane-johnson: If there is no assignment to this variable after this point, you can make this a const.

awslambda/putsnippet/index.js
```diff
@@ -0,0 +1,40 @@
...
+console.log('Loading function');
+
+const aws = require('aws-sdk');
+
+const s3 = new aws.S3({ apiVersion: '2006-03-01' });
+
+
+exports.handler = (event, context, callback) => {
+    console.log(JSON.stringify(event));
+    let user_id = event.pathParameters.user_id;
```
dane-johnson: You used camelCase for this variable in all the other functions, use it here for consistency.

PR #207, opened 2017-03-17
Parser now Handles Booleans

Hopding:
Updates the UI to handle the parser update for https://github.com/maryvilledev/codesplain/issues/17.

screen shot 2017-03-16 at 8 06 24 pm


src/util/codemirror-utils.js
```diff
@@ -9,13 +9,28 @@ function highlightNode(codeMirror, node, filters, parentColor) {
   let color = parentColor;
 
   // If we aren't ignoring this token...
-  if (ignoredRules.indexOf(node.type) === -1  && rules[node.type]) {
-    color = rules[node.type].color; // Get the color for this token's type
+// <<<<<<< HEAD
```
solkaz: Please remove evidence of your merge conflicts

dane-johnson: How about just combining the two into one filter, instead of having 2 separate filters for True and False?

dane-johnson:
How about just combining the two into one filter, instead of having 2 separate filters for True and False?

Hopding:
@dane-johnson That cannot really be done until we make use of the Parser's tagging ability, because they appear as two different types of nodes in the tree.

dane-johnson:
Can't you just make both nodes have the same key?

Hopding:
Nope, because the key _is_ the token type.

dane-johnson:
And the token type is literally just True? That's not really a token type it's the token itself.

Hopding:
It is how it appears in the tree. It normally wouldn't appear at all, because it is a terminal node. Again though, this is something that can be changed when we update the UI to use the tagging ability of the parser.

dane-johnson:
Maybe we should hold off on this merge until then. It seems silly to me to have a filter that literally just highlights one word that's already in the name of the filter.

solkaz:
It's fine for merging as is. It does not affect performance or functionality in a negative way. If we want to refactor the code, fine, we can do that later, but stylistic changes shouldn't block a PR.

dane-johnson:
☝️ I wholeheartedly disagree

PR #218, opened 2017-03-17
Regression fixes

### LINKED ISSUE #217: Fix Highlighting of filters panel and saving issues due to new parser

Hopding:
Fixes #217


__tests__/components/RuleLabel.test.jsx
```diff
@@ -15,6 +15,7 @@ describe('<RuleLabel />', () => {
         color=""
         rule=""
         count=""
+        isActive={false}
```
solkaz: There should be snapshot tests for when `isActive={true}`

PR #236, opened 2017-03-20
Add s3deploy

### LINKED ISSUE #232: Update Circle to push master branch and release tags to S3

thebho:
Closes #232

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,6 +11,6 @@ deployment:
     commands:
         - /bin/bash publishTag codesplain-ui
   master:
-    branch: [master, addS3deploy]
+    branch: master
     commands:
       - /bin/bash publish codesplain-ui-dev
--- a/circle.yml
+++ b/circle.yml
@@ -4,3 +4,10 @@ dependencies:
 test:
   override:
     - yarn test
+
+deployment:
+    aws:
+    branch: [addS3deploy, master]
+    commands:
+      - yarn
+      - /bin/bash publish.sh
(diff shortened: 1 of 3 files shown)


thebho:
I added AWS credentials to the Circle project too.

publishTag
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+BUCKET=codesplain-ui
```
dane-johnson: As far as I can tell this is the only difference between these two scripts. Make it a flag instead of 2 scripts.

circle.yml
```diff
@@ -4,3 +4,13 @@ dependencies:
 test:
   override:
     - yarn test
+
+deployment:
+  release:
+    tag: /v[0-9]+(\.[0-9]+)*/
+    commands:
+        - /bin/bash publishTag
```
dane-johnson: Explicitly calling bash here is redundant, given the shebang in the bash script.

circle.yml
```diff
@@ -4,3 +4,13 @@ dependencies:
 test:
   override:
     - yarn test
+
+deployment:
+  release:
+    tag: /v[0-9]+(\.[0-9]+)*/
+    commands:
+        - /bin/bash publishTag codesplain-ui
```
dane-johnson: This refers to a script that no longer exists

thebho:
Updated for Dane

circle.yml
dane-johnson: This should work as simply `./publish` because the shebang identifies the correct interpreter.

thebho:
Again?

PR #238, opened 2017-03-20
Added authorization function

### LINKED ISSUE #179: Add a Github OAuth Custom Authorizer Lambda Function that the Gateway API can use
1. Takes a token and a request
2. Validates Token from Github
3. Completes the request/succeeds if the token is valid

dane-johnson:
Closes #179

### DIFF
--- a/awslambda/authorizer/index.js
+++ b/awslambda/authorizer/index.js
@@ -11,25 +11,23 @@ const httpsOpts = (token) => ({
   }
 })
 
-const generatePolicy = (principleId, effect, resource) => {
-  let authResponse = {};
-
-  authResponse.principleId = principleId;
-  if (effect && resource) {
-    const policyDocument = {
-      Version: '2012-10-17', //Default Version
+function generatePolicy(principalId, effect, resource)
+{
+  return {
+    principalId,
+    policyDocument: {
+      Version: '2012-10-17',
       Statement: [{
-        Action: 'execute-api:Invoke', //Default Action
+        Action: 'execute-api:Invoke',
         Effect: effect,
         Resource: resource
       }]
-    };
-    authResponse.policyDocument = policyDocument;
-  }
-  return authResponse;
+    }
+  };
 }
 
 exports.handler = (event, context, callback) => {
+  console.log(event)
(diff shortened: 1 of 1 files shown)


awslambda/authorizer/index.js
```diff
@@ -0,0 +1,42 @@
...
+exports.handler = (event, context, callback) => {
+  const token = event.authorizationToken;
+  const opts = httpsOpts(token);
+  axios.get('https://api.github.com/', opts)
+    .then(() => {
+      callback(null, generatePolicy('user', 'Allow', event.methodArn))
+    })
+    .catch(() => {
+      callback('Unauthorized')
+    })
```
solkaz: The `catch` function should be the second argument to the `then`. https://developers.google.com/web/fundamentals/getting-started/primers/promises#promise-terminology

dane-johnson: This is the way we've always done it with axios.
https://github.com/mzabriskie/axios#example

solkaz: We've migrated away from that pattern as isn't necessarily correct. https://developers.google.com/web/fundamentals/getting-started/primers/promises#chaining

solkaz: If you look in the SnippetArea container and the Auth component, you will see that pattern being applied.

dane-johnson: From the Google article.
> You can also use catch():
> 
> `get('story.json').then(function(response) {
>   console.log("Success!", response);
> }).catch(function(error) {
>   console.log("Failed!", error);
> })`
> 
> There's nothing special about catch(), it's just sugar for then(undefined, func), but it's more readable. Note that the two code examples above do not behave the same, the latter is equivalent to:
> 
> `get('story.json').then(function(response) {
>   console.log("Success!", response);
> }).then(undefined, function(error) {
>   console.log("Failed!", error);
> })`
> 
Nothing incorrect about it. Just a way of separating success and failure.

solkaz: What I'm saying is that the two should be bundled together, as they're concerned with the same function call. But that's trivial I guess

PR #249, opened 2017-03-21
Use auth lambda 180

### LINKED ISSUE #180: The UI should hit the Gateway API and a Lambda function when a user logs in to Github
Replaces what the Express server currently does with Github Application Key/Secret

Hopding:
Fixes #180. This PR adds an AWS lambda that will take a GitHub authorization token and return an access token for the app. It also removes the `/auth` endpoint from the express backend, and points to the API instead. Note that you will need to add the `REACT_APP_API_URL` env var to your `.env` file to login with this branch.


.gitignore
```diff
@@ -18,3 +18,6 @@ yarn-error.log*
 
 # redis
 /data/redis
+
+# AWS Lambda
+/awslambda/github-access-code-getter/node_modules
```
dane-johnson: You would do well to generalize this to all directories named node_modules.

awslambda/github-access-code-getter/index.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
```
dane-johnson: The function name breaks the camelcase naming convention

server/server.js
```diff
@@ -9,19 +9,6 @@ var axios = require('axios')
 var _ = require('lodash');
 _.mixin(require('congruence'));
```
dane-johnson: Remove unnecessary imports, and remove the dependencies that are no longer needed.

dane-johnson:
So were still using Heroku for storing of snippets?

circle.yml
```diff
@@ -9,8 +9,14 @@ deployment:
   release:
     tag: /v[0-9]+(\.[0-9]+)*/
     commands:
-        - ./publish release $CIRCLE_TAG
+        - yarn build:
+            environment:
+              REACT_APP_API_URL: api.codesplain.io/prod
```
thebho: I don't think we want the prod API to be a `/prod` path.  I would say `/api` or i can look into having no path route to the prod API

circle.yml
```diff
@@ -9,8 +9,14 @@ deployment:
...
+        - ./publish codesplain-ui
```
thebho: This is a regressio

Hopding: How so? What should it be?

thebho: ` - ./publish release $CIRCLE_TAG`  It's pretty obvious in the diff

PR #250, opened 2017-03-21
Authorizes conditionally

### LINKED ISSUE #188: Limit Github Auth to only members of Maryville DevCenter or https://github.com/LaunchCodeEducation organizations

dane-johnson:
Closes #188 
- [x] Assures all requests come from a token granted to Codesplain.
- [x] Only allows logins from Maryville and Launchcode orgs
- [x] Revoke token if incorrect org


__mocks__/aws-sdk.js
```diff
@@ -0,0 +1 @@
+module.exports = {}
```
solkaz: Is Jest unable to mock this?

circle.yml
```diff
@@ -7,7 +7,11 @@ dependencies:
     - yarn
 test:
   override:
-    - yarn test
+    - yarn test:
+        environment: #Don't expose our secrets while testing
+          CLIENT_ID: birdperson
+          CLIENT_SECRET: molting
```
solkaz: :bird: :man:

__mocks__/aws-sdk.js
dane-johnson: It isn't in the dependencies, so the mock fails.

circle.yml
dane-johnson: 😆

__tests__/lambdas/authorizer.js
```diff
@@ -0,0 +1,68 @@
+const mockClientId = process.env.CLIENT_ID
+
+import { handler } from '../../awslambda/Authorize'
+
+const mockMethodArn = 'Teleport'
+
+const mockValidToken = 'ricksanchez'
+const mockInvalidToken = 'mortysmith'
+
+jest.mock('axios', () => {
```
solkaz: Use `moxios` if you need to mock axios. Check the `app` actions test file for an example

PR #262, opened 2017-03-22
Add indicator of Markdown support in AnnotationEditor

### LINKED ISSUE #139: Indicate that annotations will be rendered as Markdown
There's no indication that annotations are rendered as Markdown, we should indicate this feature to users.

solkaz:
Fixes #139 

Adds a Markdown logo and indicator that Markdown is supported in the AnnotationEditor

### DIFF
--- a/src/components/AnnotationEditor.jsx
+++ b/src/components/AnnotationEditor.jsx
@@ -8,13 +8,13 @@ import markdownRendererOptions from '../util/markdown-renderer-options';
 import markdownLogo from '../../res/markdown-logo.svg';
 
 const styles = {
-  'markdown-hint-text': {
+  'markdownHintText': {
     'color': '#d3d3d3',
   },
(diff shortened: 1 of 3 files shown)


src/components/AnnotationEditor.jsx
```diff
@@ -63,15 +77,31 @@ class AnnotationEditor extends React.Component {
...
         />
+        <a
+          href="http://commonmark.org/help"
+          style={styles['markdown-indicator']}
+          target="_blank"
+        >
+          <img
+            alt="Markdown Logo"
+            src={markdownLogo}
+            style={styles['markdown-logo']}
```
dane-johnson: I think `styles.markdown-logo` is better, this seems to indicate a key-value model.

dane-johnson: Also switch to camelcase to uphold React convention.

PR #272, opened 2017-03-24
Add navigation confirmation

### LINKED ISSUE #211: Add Alert If User Navigates Away With Unsaved Data
Right now the user could leave the tab and close it without ever being notified of unsaved data.

solkaz:
Fixes #211

If a user attempts to navigate away (i.e. close the tab/browser, go to another link) and there are unsaved changes, then they will be asked to confirm their decision


src/containers/AppBody.jsx
```diff
@@ -15,31 +15,38 @@ const API_URL = process.env.REACT_APP_API_URL;
...
-      const { dispatch } = this.props;
-      if (!username && !id) {
-        const signInState = cookie.load('signInState');
-        if (signInState) {
-          cookie.remove('signInState', { path: '/' });
-          cookie.remove('signInRedirect', { path: '/' });
-          dispatch(restoreState(signInState));
-        }
-        return null;
+    const {
```
Hopding: Could you please add a comment here, such as `// Extract variables from this.props`? This threw me off initially, as it doesn't quite look like typical destructuring syntax (to me, at least).

dane-johnson: No harm in a comment, but I'd say this is pretty recognizable.

package.json
```diff
@@ -3,13 +3,14 @@
   "version": "0.1.0",
   "private": true,
   "devDependencies": {
-    "babel-cli": "^6.23.0",
-    "babel-preset-es2015": "^6.22.0",
+    "babel-cli": "^6.24.0",
+    "babel-preset-es2015": "^6.24.0",
```
dane-johnson: I assume this doesn't break anything?

src/containers/SnippetArea.jsx
```diff
@@ -1,12 +1,13 @@
-import React, { PropTypes } from 'react';
-import { connect } from 'react-redux';
 import axios from 'axios';
-import _ from 'lodash'
+import _ from 'lodash';
 import { CardText, Snackbar } from 'material-ui';
-import { browserHistory } from 'react-router'
+import React, { PropTypes } from 'react';
```
dane-johnson: I think React should be the first import here. Makes it obvious that this is React code.

package.json
solkaz: I've ran it myself and it works fine, but you're welcome to try it yourself.

src/containers/SnippetArea.jsx
solkaz: ¯\_(ツ)_/¯

solkaz: It's not hard to read a few lines down, and I've sorted the dependencies alphabetically. That's all

dane-johnson: I think they should be sorted by order of importance. React is the most important so it should go first.

PR #276, opened 2017-03-27
Verify User Identity on POST, PUT, & DELETE of Snippets

### LINKED ISSUE #244: When calling POST/PUT/DELETE on a snippet, confirm the request came from that user
As `userA` I can't POST a snippet into `userB`'s folder.
Only `userB` can PUT or DELETE `userB`'s snippets

Hopding:
Fixes #244. This PR adds an `AuthorizeToken` AWS lambda that verifies a GitHub access token belongs to the specified user. It is used by the other lambas to confirm a request is authorized to perform DELETEs, PUTs, and POSTs on a resource (not that it is not used for GETs).


.gitignore
```diff
@@ -18,3 +18,6 @@ yarn-error.log*
 
 # redis
 /data/redis
+
+# AWS Lambda
```
dane-johnson: Unnecessary, node_modules is gitignored above.

Hopding:
They are different in that the `GithubAuthorizer` lambda only checks for a valid GitHub access token. The `AuthorizeToken` lambda (perhaps poorly named, feel free to suggest a better one) actually looks at which user the access token belongs to, and compares that with the resource being requested, and authorizes based on its findings. 

The result is that if I use a token for my account, `Hopding`, I am able to POST/PUT/DELETE to `api.codesplain.io/dev/users/Hopding/...`. But, I cannot perform those methods on `api.codesplain.io/dev/dane-johnson/...`, or any other path. This is because although I have a valid token, the token is only for my user, `Hopding`, and not your user, `dane-johnson`.

This validation is not done for GETs, because that would make it impossible to share links to snippets.

dane-johnson:
Just add that functionality to the Authorizer. Will prevent redundant code.

Hopding:
We discussed this together with Brian last week. Initially you told me you thought it belonged in the `Authorizer`. Then you said you agreed with Brian, that it should be in the second stage (not the `Authorizer`). Are you changing your mind again now?

I would prefer to leave it as is, because it does accomplish the feature request of this story. If you still feel it belongs in the `Authorizer`, then I'd prefer to see a separate story for that. However, I am not sure that moving it to the `Authorizer` makes more sense, because the verification of user identity is not something that should happen on _all_ requests; it should not occur for GETs - only on POSTs, PUTs, and DELETEs.

dane-johnson:
Forgot that convo. My bad.

PR #277, opened 2017-03-27
Fix SnippetArea to use more vertical space

### LINKED ISSUE #224: Snippet window should fill in the entire vertical space before scrolling

solkaz:
Fixes #224 

Applied CSS code to make the `SnippetArea` use more vertical space. As a nice side-effect, the CodeMirror instance will need to scroll less initially, so there should be less of an issue when the user adds more LOC.


solkaz:
I think that we will benefit from a layout change, as I had to make a quick-fix to make sure that the save and lock buttons didn't disappear when the browser width shrank.

dane-johnson:
Would like to see some pics though.

PR #281, opened 2017-03-27
Fixed yarn warning

### LINKED ISSUE #280: Fix Warning that is breaking master
```/home/ubuntu/codesplainUI/src/components/buttons/LoginButton.jsx
  20:5  warning  Prefer alt="" over role="presentation". First rule of aria is to not use aria if it can be achieved via native HTML  jsx-a11y/img-has-alt
```

Hopding:
Fixes #280. ⚔️


dane-johnson:
Yayy accessibility!

PR #283, opened 2017-03-27
Worker now loads script async from API

### LINKED ISSUE #248: Move the parser out of the webpack
Will consume language files based on the path:
`/{environment}/parsers/{language}`

dane-johnson:
closes #248


src/workers/parser.js
```diff
@@ -1,17 +1,39 @@
...
+          AST,
+          ruleCounts,
+        },
+      })
+      break;
+    }
+    case LOAD_PARSER: {
+      const parserURL  = action.payload;
+      importScripts(parserURL);
+      parser = Codesplain_parse_python3
```
solkaz: Where does `Codesplain_parse_python3` come from? I don't see where it was declared.

solkaz: And importScripts, as well.

solkaz: Nevermind, found the documentation for it. But that should be commented then, and it'd make sense to prepend `self.` to the `importScripts` call

solkaz: [Documentation for `importScripts`](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts)

PR #287, opened 2017-03-28
Apply flexbox CSS properties to AnnotationEditor child elements

### LINKED ISSUE #265: Fix the wrapping on Markdown link
Text should wrap in its own column and not under the Cancel/Save buttons

solkaz:
Fixes #265 

Applies the flex CSS properties to the Button and Markdown indicator elements in the `AnnotationEditor` component

[Flexbox reference](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes)


src/components/AnnotationEditor.jsx
```diff
@@ -8,11 +8,19 @@ import markdownRendererOptions from '../util/markdown-renderer-options';
 import markdownLogo from '../../res/markdown-logo.svg';
 
 const styles = {
+  'bottomContainer': {
+    display: 'flex',
+    'flexFlow': 'row wrap',
+  },
+  'button': {
+    flex: '1 1 auto',
```
Hopding: Would you please change these short hands to explicitly property declarations, each on their own line?

I'm not a fan of CSS shorthand notations. I nearly always have to take time to look up documentation on the shorthand if I need to modify it, because I cannot remember what it refers to (e.g. https://www.w3schools.com/cssref/css3_pr_flex.asp). I strongly prefer actually specifying each property explicitly on its own line.

solkaz:
@Hopding addressed comments

PR #292, opened 2017-03-29
Refactor SnippetArea and its children to use 💪  flexboxes

### LINKED ISSUE #291: The CodeMirror div should not overflow the codesplain app bar
pasted image at 2017_03_28 04_19 pm

solkaz:
Fixes #291 

SnippetArea and its children now use flexboxes for flow, and they look fantastic.


dane-johnson:
Are we bootstrap-independent then?

PR #295, opened 2017-03-29
Removed flex from top level component

dane-johnson:
Removes flex from toolbar container to support older browsers.


solkaz:
Closing as #298 addresses the same issue while keeping flexboxes

PR #297, opened 2017-03-30
Cleaned Up Project

### LINKED ISSUE #126: Moving dev dependencies out of normal dependencies
There are a few dependencies that can be moved to the `devDependencies` section, such as

- enzyme
- morgan (?)
- react-addons-test-utils
- react-test-renderer

Doing so would help us with our build times.

### LINKED ISSUE #261: Turn off Heroku/Redis after full cutover

Hopding:
Fixes #126. Fixes #261.

I removed unused files, folders, and dependencies. I also took out all lambda-related stuff, and added it to the `codesplain-lambdas` repo [here](https://github.com/maryvilledev/codesplain-lambdas/pull/9).


package.json
```diff
@@ -5,24 +5,20 @@
   "devDependencies": {
     "babel-cli": "^6.24.0",
     "babel-preset-es2015": "^6.24.0",
-    "worker-loader": "^0.8.0"
+    "worker-loader": "^0.8.0",
```
dane-johnson: I don't know if we can call these dev dependencies, since they are needed at build time.

solkaz: This is true; we do need it for the web worker middleware. Check middlewares/worker.js

package.json
```diff
@@ -5,24 +5,20 @@
...
-    "congruence": "^1.6.11",
-    "enzyme": "^2.7.1",
-    "enzyme-to-json": "^1.5.0",
-    "express": "^4.14.1",
     "highlight.js": "^9.10.0",
-    "ioredis": "^2.5.0",
     "lodash": "^4.17.4",
     "material-ui": "^0.17.1",
-    "morgan": "^1.8.1",
     "moxios": "^0.3.0",
```
dane-johnson: This can be considered a dev dependency

PR #298, opened 2017-03-30
Fix SnippetAreaToolbar flexbox issue for Chrome & Safari

solkaz:
Fixes the flexbox issue with Chrome and Safari.

The issue behind this bug has been documented [here](https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored)


dane-johnson:
Duplicate #295

solkaz:
@dane-johnson same issue, different solutions (and results). This allows us to use flexboxes in Chrome and Safari; your PR removes it.

dane-johnson:
We have competing solutions. If this is better, close the other PR so they are not both accepted.

PR #320, opened 2017-03-31
Refactor handleSnippetChanged to conditionally parse snippets

### LINKED ISSUE #273: Run the parser only if trimmed contents have changed
The parser is being run any time the snippet's contents change, but I think that it's unnecessary to do so if a whitespace is appended/prepended because it shouldn't change the meaning of the program. Before we send a snippet to be parsed, then, we should trim the snippet's contents and compare it to the saved snippet in the state; if they differ, then it needs to be parsed, but otherwise we do nothing.

solkaz:
Fixes #273 

`handleSnippetChanged` will not parse snippets if whitespaces are prepended/appended

### DIFF
--- a/src/containers/SnippetArea.jsx
+++ b/src/containers/SnippetArea.jsx
@@ -73,7 +73,11 @@ export class SnippetArea extends React.Component {
   handleSnippetChanged(snippetContents) {
     const { dispatch } = this.props;
     dispatch(setSnippetContents(snippetContents));
-    debouncedParseSnippetDispatch(snippetContents, dispatch)
+    // Don't parse the snippet if a whitespace character was prepended/appended
+    // So appending a newline of a snippet will not trigger a parse action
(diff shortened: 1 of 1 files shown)


src/containers/SnippetArea.jsx
```diff
@@ -73,7 +73,11 @@ export class SnippetArea extends React.Component {
   handleSnippetChanged(snippetContents) {
     const { dispatch } = this.props;
     dispatch(setSnippetContents(snippetContents));
-    debouncedParseSnippetDispatch(snippetContents, dispatch)
+    // Don't parse the snippet if a whitespace character was prepended/appended
+    // So appending a newline of a snippet will not trigger a parse action
+    if (snippetContents === snippetContents.trim()) {
```
dane-johnson: This will block the parse if they paste in code with leading/trailing whitespace.

solkaz:
@dane-johnson addressed comments

PR #328, opened 2017-04-03
Removes `/snippet/` from browser URL, with legacy support

### LINKED ISSUE #317: Remove /snippets/ from the UI routing
Probably need to re-route old snippets to new path

dane-johnson:
Closes #317

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -36,6 +36,8 @@ ReactDOM.render(
       <Router history={history}>
         <Route path="/auth"  component={Auth} />
         <Route path="/" component={App} />
+        {/* Leave this for legacy support */}
+        <Route path="/(:username)/snippets/(:id)" component={App} />
         <Route path="/(:username)/(:id)" component={App} />
       </Router>
     </Provider>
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -36,7 +36,9 @@ ReactDOM.render(
       <Router history={history}>
         <Route path="/auth"  component={Auth} />
         <Route path="/" component={App} />
+        {/* Leave this for legacy support */}
         <Route path="/(:username)/snippets/(:id)" component={App} />
```
solkaz: Could we include a [`<Redirect />`](https://github.com/ReactTraining/react-router/blob/v3/docs/guides/RouteConfiguration.md) component here to route them to the correct URL?
```jsx
<Route
  path="/(:username)/snippets/(:id)"
  render={<Redirect from="/(:username)/snippets/(:id)" to="/(:username)/(:id)" />}
/>
```

dane-johnson: Good idea. Couldn't get this exact implementation to work, but I added code to push the correct URL once the snippet has loaded.

PR #329, opened 2017-04-03
Added check for empty codemirror, pushed value to rectify

### LINKED ISSUE #220: Highlighting not occuring when returning to a page.
Upon returning to a previously filtered and annotated snippet, the UI doesn't immediately highlight.

dane-johnson:
Closes #220

### DIFF
--- a/src/components/Editor.jsx
+++ b/src/components/Editor.jsx
@@ -72,6 +72,12 @@ class Editor extends React.Component {
     } = this.state
 
     const codeMirrorInst = this.codeMirror.getCodeMirror();
+
+    // If this is the first time through, value may be set without Codemirror
+    // knowing about it. Push the value to codemirror to rectify
+    if (codeMirrorInst.getValue() === '' && value !== '') {
+      codeMirrorInst.setValue(value);
+    }
     codeMirrorInst.clearGutter('annotations');
     // eslint-disable-next-line array-callback-return
     markedLines.forEach((lineNumber) => {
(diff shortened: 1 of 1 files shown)


src/components/Editor.jsx
```diff
@@ -72,6 +72,12 @@ class Editor extends React.Component {
     } = this.state
 
     const codeMirrorInst = this.codeMirror.getCodeMirror();
+
+    // If this is the first time through, value may be set without Codemirror
+    // knowing about it. Push the value to codemirror to rectify
+    if (codeMirrorInst.getValue() === '' && value !== '') {
+      codeMirrorInst.setValue(value);
+    }
```
solkaz: Why not set this in `componentDidMount()`?

dane-johnson: Components mount from children up, so the `value` prop is not yet set when the Editor mounts, as the SnippetArea has not yet mounted and loaded its data.

solkaz: One thing that I don't like is the need to do these comparisons every time `componentDidUpdate()` is called. I do suggest, since we should only need to do this once, is to use lodash's [`once()`](https://lodash.com/docs/4.17.4#once) function, which should eliminate the need for those conditionals.

dane-johnson: Conditionals turned out to still be required, otherwise it lead to a strange bug where the first character would appear after the cursor. I nested the conditionals in a onced function, so they will only be checked the first time.

PR #332, opened 2017-04-03
Restricts saving to the owner of a snippet

### LINKED ISSUE #306: Give feedback to user when they try to Save over someone else's snippet
Currently you get a "Codesplination Saved" snackbar when the save doesn't actually process.

Save As is WAI

Disable the `Save` option when it's not your snippet
Return an error code when a Put fails

dane-johnson:
Closes #306

### DIFF
--- a/__tests__/actions/permissions.js
+++ b/__tests__/actions/permissions.js
@@ -9,7 +9,7 @@ describe("Actions: permissions", () => {
       }
       const expected = {
         type: actions.SET_PERMISSIONS,
-        payload: permissons
+        payload: permissions
       }
       expect(actions.setPermissions(permissions)).toEqual(expected)
     })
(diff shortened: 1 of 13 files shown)


dane-johnson:
Good catch. Fixed.
screen shot 2017-04-03 at 3 35 23 pm

Hopding:
That particular problem is now fixed, but another seems to have arisen:

1. Logout of the app.
2. Enter some snippet contents.
3. Login so you can save.
4. Notice that despite being logged in, you cannot save. You can only "Save As".

### Step 1
screen shot 2017-04-03 at 3 40 28 pm

### Step 4
screen shot 2017-04-03 at 3 38 35 pm

dane-johnson:
QA level 9000

PR #335, opened 2017-04-03
Add invalidation

thebho:
Connects #294


thebho:
Salt accepted

thebho:
It's a circle change, so there is a 94% chance it will fail the first time

Issue #310, opened 2017-03-30
Support multiple languages in the language selector


thebho:
- Python3, Java
- Pull new parser when you chose the language
- Confirm that the parsers are being cached

dane-johnson:
This will involve changing some hard-coded values, and standardizing the name the environment variable is exported under in [Codesplain](https://github.com/maryvilledev/codesplain)

Hopding:
@dane-johnson What environment variable are you referring to?

dane-johnson:
Awaiting #341 to avoid conflicts.

PR #341, opened 2017-04-04
Language switch

### LINKED ISSUE #312: When you change a language, the code mirror language needs to change

solkaz:
## Description

- Add `LanguageSelector` component (wrapper around `SelectField`)
- Adds the `language` property to the `app` state
- `SnippetArea` will receive this prop in `mapStateToProps` and pass it to the `Editor` and `SnippetAreaToolbar`
- `Editor` will be able to get the necessary files for that language mode
- `SnippetAreaToolbar` will set the `LanguageSelector`'s value to the language
- Selecting a different language will dispatch an action to set the new language (note that the SelectField is still disabled)
- Refactor tests to use `defaultProps` to cut down verbosity

## Motivation and Context

Fixes #312 
This PR does **not** implement #310. It will not fetch the parser files for a language if you change it, and the SelectField is still disabled, so you cannot change languages anyways. This simply implements much of the logic necessary for #310.

## Checklist:


__tests__/components/Editor.test.jsx
```diff
@@ -5,21 +5,24 @@ import getMuiTheme from 'material-ui/styles/getMuiTheme';
 
 import Editor from '../../src/components/Editor';
 
+const defaultProps = {
```
dane-johnson: Good idea here.

__tests__/reducers/app.js
```diff
@@ -8,6 +8,7 @@ describe('Reducer: App', () => {
       AST: {},
       filters: {},
       hasUnsavedChanges: false,
+      language: 'python',
```
dane-johnson: We might need to save parser lang and Codemirror lang seperately, as the parser we need to pull is `python3`

src/components/LanguageSelector.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <SelectField
+      disabled
+      id="languageSelector"
+      onChange={onChange}
+      style={style}
+      value={language}
+    >
+      <MenuItem value="python" primaryText="Python 3"/>
```
dane-johnson: Again, this will work for setting Codemirror mode correctly, but if we want to use this as a variable to retrieve a parser, I'd rather have to shave off the `3` for Codemirror than clear up an ambiguous call to the parser.

solkaz:
I don't see why the parser name can't match the CodeMirror mode. That reduces having to have to fields for a language (one for the parser name, one for CodeMirror)

dane-johnson:
Then we've named our parsers based on the names of the Codemirror modes. This will prevent adding parsers for unsupported Codemirror languages, and restricting to one parser for languages that use the same Codemirror mode (i.e. Python, Lisps).

solkaz:
I don't believe that having multiple parsers for a single CodeMirror language or supporting non-CodeMirror-supported languages has ever been a goal, but I can see that point.

dane-johnson:
I think that now, since we only support two languages and all the tokens for the languages are still hard-coded in the repo, it wouldn't be wrong to write a util function that takes the parser language and converts it to a Codemirror mode.

solkaz:
We had one, then it got taken out 🤔 but I'm adding it right now.

solkaz:
@dane-johnson addressed comments

dane-johnson:
Looks good, but snippets should now refer to the parser language instead of the Codemirror language, which is `python3`

solkaz:
I don't understand your concern, because I'm already doing that.

dane-johnson:
Take a look in the reducer.

solkaz:
Please be exact and point out the offending code. All I can see is that the parser name is stored it in `snippetLanguage`, which, if you're concerned about the naming, isn't a valid reason to block.

dane-johnson:
It is the same code commented on before.

solkaz:
@dane-johnson fixed.

thebho:
Just so I'm up to date:
What is the key when changing the language selector?
What is the the language key are we saving the snippet?
What is the key for code mirror?
What is the key for the parser? 
Can they be combined into matching keys?

dane-johnson:
Should be 
1. A pretty name mapped to the parser name
2. The parser name
3. The Codemirror mode name
4. The parser name
5. No, unless we want to limit ourselves to one and only one parser for each Codemirror mode (see above)

src/containers/SnippetArea.jsx
```diff
@@ -281,6 +293,7 @@ const mapStateToProps = state => ({
   annotations: state.app.annotations,
   AST: state.app.AST,
   filters: state.app.filters,
+  language: state.app.language,
```
dane-johnson: this needs to be `state.app.snippetLanguage`

PR #347, opened 2017-04-04
Fix bug that flags users having unsaved changes when loading a snippet

### LINKED ISSUE #346: Users are incorrectly asked to confirm navigation if reloading pre-saved snippet
## Expected Behavior
User should not be asked to confirm navigation away from Codesplain if they reload a pre-saved snippet and do not edit anything

## Current Behavior
They will be asked to confirm navigation away from the page

1. Return to a pre-saved snippet
2. Attempt to navigate away

## Possible Solution
It's an issue with the app reducer as it sets the `hasUnsavedChanges` variable to true

solkaz:
## Description

Edits the `app` reducer to set `hasUnsavedChanges` to false when reloading a snippet

## Motivation and Context
Fixes #346 

Necessary because users will be asked if they wish to navigate away even though they haven't made any changes, which is a minor annoyance

## Checklist:

### DIFF
--- a/__tests__/reducers/app.js
+++ b/__tests__/reducers/app.js
@@ -190,7 +190,11 @@ describe('Reducer: App', () => {
       type: actions.RESTORE_STATE,
       payload: savedState
     };
-    expect(reducer(undefined, action)).toEqual(savedState);
+    const expected = {
+      ...savedState,
+      hasUnsavedChanges: false,
+    };
+    expect(reducer(undefined, action)).toEqual(expected);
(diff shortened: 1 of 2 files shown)


dane-johnson:
Is the logic to check for this flag already in place?

dane-johnson:
Found it. 
https://github.com/maryvilledev/codesplainUI/blob/e38931147e96c7e59f02ffa33e1800ed57327589/src/index.js#L29

PR #351, opened 2017-04-05
Fix routing bug when saving new snippet

### LINKED ISSUE #350: Pre-saved snippets are not actually being updated
## Expected Behavior
If you have already saved a snippet and click the "Save" button (not Save As), it will make a PUT request to update the snippet

## Current Behavior
It sends a POST request every time you click save. If you refresh the page, React-Router will have its history updated, so it can get the id and then properly do PUT requests, but this is before the user has done so

1. Start with a new snippet
2. Enter a title and body
3. Click save
4. Click save again. If you look at your network tab in the developer tools, two POST requests will have been sent, when there should've been a POST and then a PUT

## Possible Solution
This is due to how we are handling browser changes because we are using `browserHistory` instead of using the `withRouter` HOC. When you push to `browserHistory`, the app's router isn't in sync so it won't be able to read the id of the snippet in `handleSave()`

The need for these changes was partially discussed in #342

solkaz:
## Description
Utilizes React-Router's `withRouter` to allow components to properly manage the browser history.

## Motivation and Context
Fixes #350
This is necessary because we previously used `window.location = '...'` or `browserHistory.push('...')` to update the URL, which _will_ change the URL in the browser, but React-Router will be unaware of these changes. By using the `withRouter` HOC, we allow our components to access the params and browser history (because they are passed as props) so that they can modify the history and keep React-Router in sync.

## Checklist:


solkaz:
Not sure that there's a feasible testing strategy with React-Router v3. v4 seems to offer utilities that could help with testing, but for now, I don't know how I would be able to test this.

thebho:
Merge Conflicts

solkaz:
@thebho addressed conflicts

PR #352, opened 2017-04-05
Shows a different error message based upon the reason for login rejection

dane-johnson:
## Description

Users will now receive distinct error messages based on the response code
400: `Failed to login with GitHub, sorry.`
403: `Sorry, you are not a member of an organization authorized to use this application.`
Additionally, refactored http response code.

## Motivation and Context

Closes maryvilledev/codesplain#42

## Checklist:

### DIFF
--- a/src/components/Auth.jsx
+++ b/src/components/Auth.jsx
@@ -12,6 +12,19 @@ export const errors = {
   " this application."
 }
 
+const resolveErrorMessage = (status) => {
+  switch (status) {
+    case 403: {
+      return errors.badOrg;
+    }
+    // eslint-disable-next-line
+    case 400: //Intentional fallthrough
+    default: {
+      return errors.badCode;
+    }
+  }
+}
+
 /*
 <Auth /> is the component that is rendered for the '{{url}}/auth' endpoint (which
 is the endpoint the user is sent to after signing in to GitHub). It initially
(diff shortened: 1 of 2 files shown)


src/components/Auth.jsx
```diff
@@ -33,27 +38,40 @@ class Auth extends React.Component {
...
-        // Can pull lots of other stuff out of res.data if needed
-        const { login, avatar_url } = res.data;
-        cookie.save('userAvatarURL', avatar_url, { path: '/' });
-        cookie.save('username', login, { path: '/' });
-        this.setState({ waiting: false });
-      }, err => {
-        // Failed to pull in user info, but that's fine. Log and continue.
-        this.setState({ waiting: false });
+      .catch( err => {
+        // If this fails, we need to make sure the error dialog shows
```
solkaz: Could we possibly extract this into a helper function?

src/components/Auth.jsx
```diff
@@ -33,27 +38,40 @@ class Auth extends React.Component {
         cookie.save('token', token, { path: '/' });
 
         // Return Promise to get the user's basic info
-        return axios.get('https://api.github.com/user', {
+        axios.get('https://api.github.com/user', {
```
solkaz: Why take out the return?

dane-johnson: Regression, unintentional.

PR #353, opened 2017-04-06
Refactor project to use dispatched save actions

### LINKED ISSUE #271: Refactor saving methods
There is a redux action to save, and there are also `handleSave` and `handleSaveAs` functions on the `SnippetArea` container. There don't need to be that many functions, and it shouldn't be in the container; the handler functions should instead dispatch an action to Save or Save As

solkaz:
## Description
This PR refactors the project to use dispatched Redux actions to save the state. Container components do not make the save requests themselves but instead dispatch an action to do so.

## Motivation and Context
Fixes #271 
Previously, container components that needed to make a request to save the state would simply do it themselves, which meant that if the logic behind saving changed, then there would be multiple places in the project that would need to be refactored. Instead, there are now two redux actions, `saveNew()` and `saveExisting()`, that can be dispatched to save the state to AWS. 

## Checklist:

### DIFF
--- a/src/util/requests.js
+++ b/src/util/requests.js
@@ -1,0 +1,13 @@
+const API_URL = process.env.REACT_APP_API_URL;
+
+/* Construct the endpoint to make a REST request to. Only the username is
+ * field is required; the snippetId will be left blank for POST requests,
+ * and can be supplied when needed (say, for PUT requests).
+ *
+ */
+export const makeSaveEndpointUrl = (username, snippetId = '') => {
+  if (snippetId) {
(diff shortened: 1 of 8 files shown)


src/util/requests.js
```diff
@@ -0,0 +1,14 @@
+const API_URL = process.env.REACT_APP_API_URL;
+
+/* Construct the endpoint to make a REST request to. Only the username is
+ * field is required; the snippetId will be left blank for POST requests,
+ * and can be supplied when needed (say, for PUT requests).
+ */
+export const makeSaveEndpointUrl = (username, snippetId = '') => {
```
dane-johnson: Smart

PR #361, opened 2017-04-06
Prevent deprecated filters from being displayed

### LINKED ISSUE #358: If a filter is in the `ignore rules` array, don't display it in the Filters section, even if they have saved it to S3
Right now the `Atoms` filter will still display on previously saved snippets

## Expected Behavior
All deprecated filters should be ignored

## Current Behavior
Deprecated filters are displayed, but there is no functionality.

1. Load a snippet saved before 4/6
2. If that snippet has any `Atoms`, you can see that they are still displayed

## Possible Solution
Ignore those filters before rendering.
Create a batch lambda that can migrate the saved snippets to follow the new standards

solkaz:
## Description
This PR adds two util functions: `removeDeprecatedFilters` and `removeDeprecatedFiltersFromState`, the former removing deprecated filters from a filter object, and the latter calling the former to remove deprecated filters from a state object. `removeDeprecatedFiltersFromState` is called when loading a snippet from S3, so when the `restoreState` action is dispatched, the state object that is restored does not contain any deprecated filters. When the user saves afterwards, the deprecated filters will not be in the state object so they cease to exist (in that snippet)

## Motivation and Context
Fixes #358 

## Checklist:


thebho:
👀

thebho:
Conflicts and needs more reviewers

PR #368, opened 2017-04-07
Loads the parser rules async from API

### LINKED ISSUE #311: Figure out how to handle Rules mapping for a new language
Do we add another layer to the map or pull in a new rules file based on language

dane-johnson:
## Description

Rules are now loaded on the parser request, and are cached locally similarly to the parser.

## Motivation and Context

Closes #311 

## Checklist:


src/util/rules.js
```diff
@@ -1,60 +1,65 @@
...
+          color: row[3],
+        };
+        return map;
+      }
+      const rules = rows.filter((row) => row[1] === '1')
+        .reduce(reducer, {});
+      const ignoredRules = rows.filter((row) => row[1] === '0')
+        .map(row => row[0]);
+      return {rules, ignoredRules}
+    });
```
solkaz: Why don't we use an external library to parse CSVs?

dane-johnson: Parsing CSVs is a cakewalk and I didn't feel the need to learn a new API.

solkaz: The point is that that is less code that we have to maintain. Did you write tests to ensure you properly parse CSVs? What if there's a major change to file structure? Using an external library abstracts these things so that we don't have to think about these things.

dane-johnson: CSVs have standard structure, and I'm not worried about that changing any time soon. I decoupled the parsing function and added a test.

thebho:
Is this happy now?

solkaz:
No, because there is state management that's occurring in a util file when it essentially should be handled by a reducer.

dane-johnson:
The state management is happening in the reducer.

solkaz:
Subscribing to the store is unnecessary because setting the rules can be done in a reducer function to manage the parser state. You're setting the rules/ignored rules, and the cache, in free variables in util files, when the point of util files is to have small helper functions (not state managers). Instead, you could easily dispatch actions to update the rules state, and especially if you chain it with your `loadParser` function (if it's a Promise)

dane-johnson:
Maybe the files should be renamed, but they need to know the rules. Caching is just a side-effect, and shouldn't be part of the functional reducers.

solkaz:
What needs to know the rules? And why wouldn't it/they not be able to access the rules if it was stored in a reducer, when we're already able to do so from other reducers? And caching should absolutely be handled by the reducer.

dane-johnson:
Disagree. Reducers are not a catchall, they are just meant to facilitate transition from one application state to another.

solkaz:
That's what you're doing right now. You modify the application's (or part of it) state when you set the rules and ignored rules.

dane-johnson:
No the state is being handled by the reducer. When the the rules module is notified of a change, it modifies the highlighter to use new rules.

Hopding:
@solkaz What do you mean "clean the test files"?

src/workers/parser.js
```diff
@@ -29,7 +32,7 @@ self.onmessage = ({ data : action }) => {
         parser = parserCache[language];
       } else {
         // Load and execute script from the URL
-        self.importScripts(`${API_URL}/parsers/${language}`);
+        self.importScripts(makeParserEndpointUrl(language));
```
dane-johnson: Did this really make it easier to read?

PR #371, opened 2017-04-10
Reset application state when navigating to home page

### LINKED ISSUE #366: Codesplain does not reset when you click the logo
## Expected Behavior
If I click the logo, it should route me to the home view, with all fields cleared out

## Current Behavior
Data from the previous snippet still persists

1. Make a new snippet, save it
2. Click the logo
3. Notice that the browser location did change, but that the data from the previous snippet persists

## Possible Solution
I believe that this is occurring because when the route changes, `App`'s state doesn't change and therefore doesn't cause a re-render

solkaz:
## Description
This PR properly resets the application state when the Codesplain logo is clicked so that the user is presented with a clean, new view. There is a check added before wiping any changes, so that if there are unsaved changes, the user will be asked to confirm navigation before doing so, and will reset state/redirect them if there are no unsaved changes.

## Motivation and Context
Fixes #366 

## Checklist:

### DIFF
--- a/src/containers/CodesplainAppBar.jsx
+++ b/src/containers/CodesplainAppBar.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import AppBar from 'material-ui/AppBar';
+import Dialog from 'material-ui/Dialog';
+import FlatButton from 'material-ui/FlatButton';
 import { withRouter } from 'react-router';
 import cookie from 'react-cookie';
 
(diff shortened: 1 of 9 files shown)


src/containers/CodesplainAppBar.jsx
```diff
@@ -1,9 +1,12 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import AppBar from 'material-ui/AppBar';
+import Dialog from 'material-ui/Dialog';
+import FlatButton from 'material-ui/FlatButton';
```
dane-johnson: Consider doing all these imports as one with destructuring

PR #383, opened 2017-04-12
Display Selectable List of a User's Snippets

### LINKED ISSUE #117: If a user is authenticated from Github, show all of their saved snippets

Hopding:
## Description

This PR adds a list containing all of a user's snippets to the menu in the title bar. If a user selects a snippet from the list, that snippet is then displayed.

## Motivation and Context

Fixes #117. This change is needed because without it, a user has not way to view all of their snippets.

## Checklist:

screen shot 2017-04-11 at 7 25 39 pm


dane-johnson:
The arrow should point at the snippets.

thebho:
I don't think the arrow on the right is worth fighting.  That is expected behavior of a submenu on the right side of a screen

__tests__/actions/app.js
```diff
@@ -192,6 +192,7 @@ describe('Actions: App', () => {
     afterEach(() => {
       cookie.remove('token');
       cookie.remove('username');
+      cookie.remove('token');
```
solkaz: Extraneous?

src/actions/user.js
```diff
@@ -0,0 +1,54 @@
...
+export const updateUserSnippetsFailed = () => ({
+  type: UPDATE_USER_SNIPPETS_FAILED,
+});
+
+export const updateUserSnippets = () => {
+  return (dispatch) => {
+    // Load requisite cookies, return Promise if they aren't present
+    const token    = cookie.load('token');
+    const username = cookie.load('username');
+    if (!token || !username) {
```
solkaz: Don't see why this is necessary, because AFAIK you aren't chaining promises on your calls to this function. Moreover, you can just let the request fail and the catch statements will handle it.

src/actions/user.js
```diff
@@ -0,0 +1,54 @@
...
+      Authorization: `token ${token}`,
+    };
+    dispatch(updateUserSnippetsStarted());
+    return axios.get(makeSaveEndpointUrl(username), { headers })
+      .then(
+        res => {
+          dispatch(setUserSnippets(res.data));
+          dispatch(updateUserSnippetsSucceeded());
+        },
+        err => dispatch(updateUserSnippetsFailed)
```
solkaz: You're not invoking the action creator. Should be
```js
err => dispatch(updateUserSnippetsFailed())
```

src/components/menus/SnippetList.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { MenuItem } from 'material-ui';
+import ArrowDropRight from 'material-ui/svg-icons/navigation-arrow-drop-right';
+
+const makeMenuItems = (userSnippets, onClick) => {
+  if (!Object.keys(userSnippets).length === 0) return null;
+  return Object.keys(userSnippets).map(key => (
+    <MenuItem
+      key={key}
+      primaryText={userSnippets[key].snippetTitle}
+      onClick={() => onClick(key)}
```
solkaz: If possible, let's try to not use arrow functions for event handlers.

src/components/menus/SnippetList.jsx
```diff
@@ -0,0 +1,32 @@
...
+    />
+  ));
+};
+
+const SnippetList = ({ titles, onClick }) => {
+  return (
+    <MenuItem
+      primaryText="My Snippets"
+      rightIcon={<ArrowDropRight />}
+      menuItems={makeMenuItems(titles, onClick)}
```
solkaz: We might like to use [`_.memoize`](https://lodash.com/docs/4.17.4#memoize) here

src/containers/CodesplainAppBar.jsx
```diff
@@ -49,6 +57,11 @@ export class CodesplainAppBar extends React.Component {
     location.reload();
   }
 
+  handleSnippetSelected(key) {
+    const username  = cookie.load('username');
+    window.location = `/${username}/${key}`;
```
solkaz: Prefer using `this.props.router` to facilitate routing within the application

solkaz: So: 
```js
const { router } = this.props;
router.push(`/${username}/${key}`);
```

src/components/menus/SnippetList.jsx
Hopding: Please give an example of what you would prefer this to be.

solkaz: Essentially you would just pass the onClick prop by itself, and the callback function would take the key (or whatever value it accepts). I'm not going to block on this, though.

src/containers/CodesplainAppBar.jsx
solkaz: You would want to take out 
```js
window.location = `/${username}/${key}`;
```

Hopding: I tried removing that line, and doing so causes the app to no longer actually load the selected snippet, despite updating the URL. I'm happy to remove the line, but I will need an alternative way to load the snippet if I do.

solkaz: Hm. Then I guess it's find to leave it as-is

__tests__/actions/user.js
```diff
@@ -0,0 +1,95 @@
+import moxios from 'moxios';
```
dane-johnson: You are doing a lot of async calls and not waiting them to complete. You should set the second param of each `it` to a callback that will be completed after each test

solkaz: The tests are valid and are what I have used. There isn't anything wrong with them.

dane-johnson:
Added callback so test will await promise fulfillment

PR #412, opened 2017-04-17
Handle users who have no index.json

### LINKED ISSUE #411: Handle users with no index.json
## Expected Behavior

When a user logs in, and they do not have an `index.json` file in their S3 bucket (or they are new, and have no S3 bucket at all), the snippet list option in the app menu should not expand (or should be disabled).

## Current Behavior

Currently, if a user logs in and they do not have an `index.json` file, the following XHR request is performed:

screen shot 2017-04-17 at 4 39 43 pm

And when they expand their "My Snippets" option, they get an infinitely long empty list:

screen shot 2017-04-17 at 4 43 50 pm

This is because the axios call that requests their `index.json` file does not actually fail, as one would expect: https://github.com/maryvilledev/codesplainUI/blob/master/src/actions/user.js#L48. Instead, it succeeds with a 200 status code, but the body of the response is:
```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Key>Hopding/index.json</Key><RequestId>D4B75D9E07E90D9B</RequestId><HostId>227x9z160TTI1aRFELNU115TsDji6/FLuYidrd5hsROou0eA4D3QJBz0BVvw6Ra0n9vgIg1x/NE=</HostId></Error>
```

This is not properly handled anywhere, and seems to be causing the infinite empty list.

1. Remove your `index.json` file from the dev S3 bucket.
2. Log out of Codesplain.
3. Log in to Codesplain.
4. Expand your "My Snippets" list.

## Possible Solution

Perhaps search the response body for `<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message>`, and throw an error if it exists in the body, thus jumping into the `catch` statement of the Promise chain.

Hopding:
## Description

This PR causes an error to be thrown if a GET for a user's `index.json` returns a response containing this xml:
```xml
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message>
```
and the user's snippet list menu option is disabled until they possess snippet(s) & an index file.

## Motivation and Context

Fixes #411. This PR solves the problem of users with no `index.json` getting an infinite empty list of snippets.

## Checklist:

### User with no snippets
screen shot 2017-04-17 at 6 07 29 pm

### User with snippets
screen shot 2017-04-17 at 6 08 01 pm

### DIFF
--- a/src/actions/user.js
+++ b/src/actions/user.js
@@ -43,8 +43,8 @@ export const updateUserSnippets = () => {
       .then(res => {
         // Jump to catch block if the user has no index.json file:
         if (res.data.includes(
-          '<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message>')
-        ) {
+          '<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message>'
+        )) {
           throw new Error(`index.json does not exist for ${username}!`);
         }
 
--- a/src/actions/user.js
+++ b/src/actions/user.js
@@ -40,12 +40,20 @@ export const updateUserSnippets = () => {
     };
     dispatch(updateUserSnippetsStarted());
     return axios.get(makeSaveEndpointUrl(username), { headers })
-      .then(
-        res => {
-          dispatch(setUserSnippets(res.data));
-          dispatch(updateUserSnippetsSucceeded());
-        },
-        err => dispatch(updateUserSnippetsFailed())
-      );
+      .then(res => {
+        // Jump to catch block if the user has no index.json file:
+        if (res.data.includes(
+          '<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message>')
+        ) {
+          throw new Error(`index.json does not exist for ${username}!`);
+        }
+
+        // Otherwise they have an index.json, so update their snippets
+        dispatch(setUserSnippets(res.data));
+        dispatch(updateUserSnippetsSucceeded());
+      })
+      .catch(err => {
+        dispatch(updateUserSnippetsFailed())
+      });
   };
 };
--- a/src/actions/user.js
+++ b/src/actions/user.js
@@ -42,7 +42,8 @@ export const updateUserSnippets = () => {
     return axios.get(makeSaveEndpointUrl(username), { headers })
       .then(res => {
         // Jump to catch block if the user has no index.json file:
-        if (res.data.includes(
+        if (typeof(res.data) === 'string' &&
+            res.data.includes(
           '<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message>'
         )) {
           throw new Error(`index.json does not exist for ${username}!`);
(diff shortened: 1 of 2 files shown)


src/actions/user.js
```diff
@@ -40,12 +40,21 @@ export const updateUserSnippets = () => {
...
-          dispatch(updateUserSnippetsSucceeded());
-        },
-        err => dispatch(updateUserSnippetsFailed())
-      );
+      .then(res => {
+        // Jump to catch block if the user has no index.json file:
+        if (typeof(res.data) === 'string' &&
+            res.data.includes(
+          '<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message>'
+        )) {
```
solkaz: Consider making a utility function for this? It's a lot to fit into a conditional statement

src/components/menus/SnippetList.jsx
```diff
@@ -16,11 +16,13 @@ const makeMenuItems = _.memoize((userSnippets, onClick) => {
 });
 
 const SnippetList = ({ titles, onClick }) => {
+  const menuItems = makeMenuItems(titles, onClick);
   return (
     <MenuItem
       primaryText="My Snippets"
       rightIcon={<ArrowDropRight />}
-      menuItems={makeMenuItems(titles, onClick)}
+      disabled={menuItems ? false : true}
```
solkaz: You can use `disabled={!Boolean(menuItems)}`

src/actions/user.js
dane-johnson: Also this is xml, which should be parseable. Shouldn't need to look at it raw.

src/components/menus/SnippetList.jsx
dane-johnson: I think this is more readable as-is.

Hopding: I also find it more readable this way. I'm willing to change it though, if you can explain that what you propose is objectively better.

src/actions/user.js
Hopding: I will move it into its own function. 

I do not think we would benefit from parsing the XML, it would just be more complexity and overhead.

src/components/menus/SnippetList.jsx
dane-johnson: `objectively better` 😆

solkaz: I prefer it because it explicitly sets it to a Boolean value instead of relying on JS' truthiness. Again, it's not a reason to block, just a nitpick I notice.

src/actions/user.js
dane-johnson: Maybe just do a regex search for the specific value then

solkaz: We don't have a need to parse the XML. The only reason we receive it is when we make a request to get the index.json file and the id/username is undefined. A better solution would probably be to not make the request if either username or id is undefined

solkaz: @Hopding I know I veered you towards this direction (you previously had those checks in place), so you can lambast me when I get in today

dane-johnson: We get it if there's no index file.

dane-johnson: `res.data.indexOf('NoSuchKey') > 0` might be better

PR #416, opened 2017-04-18
Add eslint to project

### LINKED ISSUE #340: Add Eslint to the project
## Expected Behavior

It should flag style errors and bad practices.

## Current Behavior

We don't have feedback for our code style other than what create-react-app's `.eslintrc` has defined.

solkaz:
## Description
Adds eslint to the project

## Motivation and Context
Fixes #340

## Checklist:


solkaz:
For those who use Atom, I recommend you install: 
- linter
- linter-eslint

Hopding:
Merge conflicts 😮

thebho:
You only changed 96 files, I thought I asked for atleast 100 file changes for each PR

__tests__/actions/user.js
```diff
@@ -7,8 +7,8 @@ import * as actions from '../../src/actions/user';
 
 describe('Actions: User', () => {
   describe('async actions', () => {
-    const middlewares = [ thunk ];
-    const mockStore   = configureMockStore(middlewares);
```
Hopding: Does eslint not like columnar code alignment?

__tests__/containers/SnippetArea.test.jsx
```diff
@@ -1,34 +1,33 @@
...
-import { SnippetArea } from '../../src/containers/SnippetArea'
+import { SnippetArea } from '../../src/containers/SnippetArea';
 
-const mockDispatch = jest.fn()
+const mockDispatch = jest.fn();
 const mockAnnotations = {
-  "2": {
-    annotation: "They're just robots!",
+  2: {
+    annotation: 'They\'re just robots!',
```
Hopding: Maybe use backticks here ` `` ` to avoid the ugly escape?

__tests__/containers/SnippetArea.test.jsx
```diff
@@ -1,34 +1,33 @@
...
 const mockPermissions = {
   canRead: true,
-  canEdit: true
-}
-const mockSnippet = "print('Squanch ya later!')"
-const mockSnippetTitle = 'Squanchy AI'
-const mockOpenLine = 2
+  canEdit: true,
+};
+const mockSnippet = 'print(\'Squanch ya later!\')';
```
Hopding: Backticks? ` `` `

__tests__/containers/SnippetArea.test.jsx
```diff
@@ -42,31 +41,31 @@ describe('<SnippetArea />', () => {
...
+      readOnly
       snippetLanguage="python3"
       permissions={mockPermissions}
-    />
-  )
+    />,
+  );
   it('updates the snippet on handleSnippetChanged', () => {
-    const snippet = "print('Can I squanch in your room?')";
+    const snippet = 'print(\'Can I squanch in your room?\')';
```
Hopding: I <3 backticks ` `` `

__tests__/actions/user.js
solkaz: Nope.

src/actions/user.js
```diff
@@ -27,25 +27,27 @@ export const updateUserSnippetsFailed = () => ({
...
+    Accept: 'application/json',
+    Authorization: `token ${token}`,
+  };
+  dispatch(updateUserSnippetsStarted());
+  return axios.get(makeSaveEndpointUrl(username), { headers })
       .then(
-        res => {
+        (res) => {
+          // TODO: Don't execute the request if there is no username
+          if (typeof res.data === 'string') {
```
Hopding: Can you drop a comment here explaining what has happened if `res.data` is a `string`? Right now it is not at all clear why we would return if that happens.

src/components/Editor.jsx
```diff
@@ -43,66 +43,67 @@ const makeMarker = () => {
 };
 
 const pushValueToCodeMirror = _.once((value, codeMirrorInst) => {
-  if (codeMirrorInst.getValue() === '' && value !== '') {
+  if (codeMirrorInst.getValue() && value) {
```
Hopding: I think you want to flip the value of `codeMirrorInst.getValue()`.

E.g. `codeMirrorInst.getValue() === ''` => `!codeMirrorInst.getValue()`

src/components/RulesSelector.jsx
```diff
@@ -1,8 +1,8 @@
 import React, { PropTypes } from 'react';
 import RuleLabel from './RuleLabel';
 
-const makeListItems = (filters, onRuleSelected) => {
-  return Object.keys(filters)
+const makeListItems = (filters, onRuleSelected) =>
```
Hopding: I have no problem with removing the `{}`, but I think it would be best to replace them with `()`. Otherwise it can be slightly jarring to figure out the start/end of the block, since it does not all fit on one line.

src/containers/CodesplainAppBar.jsx
```diff
@@ -131,20 +132,18 @@ export class CodesplainAppBar extends React.Component {
...
         onTouchTap={this.handleConfirmNavigation}
-      />
+      />,
     ];
 
-    const { userSnippets } = this.props.userState;
+    const { userSnippets } = this.props;
     const rightElement = this.state.isLoggedIn ?
-      <AppMenu
+      (<AppMenu
```
Hopding: Are these `()` an eslint thing? If not, I don't see why we need to add them, as the `<>` delimit the JSX pretty well.

src/containers/CodesplainAppBar.jsx
```diff
@@ -168,10 +167,28 @@ export class CodesplainAppBar extends React.Component {
...
+  hasUnsavedChanges: PropTypes.bool.isRequired,
+  userSnippets: CustomPropTypes.snippets,
+};
+
+CodesplainAppBar.defaultProps = {
+  userSnippets: {},
+};
+
+const mapStateToProps = (state) => {
+  console.log('state', state);
```
Hopding: Can this log be removed?

src/containers/FilterArea.jsx
```diff
@@ -1,53 +1,53 @@
...
   handleSelectAllFilters() {
     const { dispatch } = this.props;
     dispatch(selectAllFilters());
   }
+
   render() {
     const { filters } = this.props;
     const filterAreaActions = Object.keys(filters).length ?
-      <FilterAreaActions
+      (<FilterAreaActions
```
Hopding: Again, not a fan of these `()` unless eslint wants them to be there.

src/containers/SnippetArea.jsx
```diff
@@ -145,20 +146,20 @@ export class SnippetArea extends React.Component {
     }
      // Update a pre-existing snippet
     if (id) {
-      return dispatch(saveExisting())
+      dispatch(saveExisting())
```
Hopding: Since you removed the `return`s, the logic is no longer halted at those points. So I believe you will need to either add them back or else wrap wrap the logic starting on line `157` in an `else`.

src/index.js
```diff
@@ -3,14 +3,16 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
 import injectTapEventPlugin from 'react-tap-event-plugin';
+/* eslint-disable */
```
Hopding: Why disable here?

src/middlewares/worker.js
```diff
@@ -1,7 +1,9 @@
 import createWorkerMiddleware from 'redux-worker-middleware';
 
-/* eslint-disable import/no-webpack-loader-syntax */
+/* eslint-disable */
```
Hopding: Would it not be better to only disable `import/no-webpack-loader-syntax`, rather than all of eslint?

src/containers/CodesplainAppBar.jsx
solkaz: I did remove it whilst merging

src/containers/CodesplainAppBar.jsx
solkaz: If you have a multi-line JSX tag then you need to use parentheses

src/components/RulesSelector.jsx
solkaz: If you omit curly braces then it will be interpreted as returning an expression, or basically a single statement. It is an eslint thing, though

src/index.js
solkaz: It throws an error for not having the history package in the dependencies. It comes with React-Router though, so I don't see a need to specify it

src/middlewares/worker.js
solkaz: There were multiple errors associated with the proceeding line (associated with dependencies); it doesn't affect program flow and isn't something to worry about

src/util/markdown-renderer-options.js
```diff
@@ -2,14 +2,14 @@ import hljs from 'highlight.js';
 
 export default {
   langPrefix: 'hljs language-',
-  highlight: function (str, lang) {
+  highlight(str, lang) {
```
Hopding: Not something I would block on, but I find this syntax more readable:
```javascript
{
...,
highlight: () => { ... },
}
```

solkaz: That was actually a bug. Good catch

src/util/parser-error-logger.js
```diff
@@ -1,14 +1,13 @@
...
+  NODE_ENV === 'development' && REACT_APP_ENABLE_PARSER_DEBUGGING === 'true';
 
-if (NODE_ENV === 'development' && REACT_APP_ENABLE_PARSER_DEBUGGING === 'true') {
-  errorFn = (err) => { console.error(err); };
-} else {
-  errorFn = () => {};
-}
+const errorFn =
+  /* eslint-disable no-console */
+  isDebuggingEnabled ? (err) => { console.error(err); } : () => {};
```
Hopding: Since we use this syntax for all the other anonymous functions, wouldn't this be preferable?
`(err) => { console.error(err); }` => `err => console.error(err)`

__tests__/actions/user.js
Hopding: That is silly :)

src/components/RulesSelector.jsx
Hopding: I realize that, but you can still use `()` instead to indicate that it is all a block belonging to the function.

src/util/parser-error-logger.js
solkaz: No, because we're not returning the result of an expression. Same applies for `ref`s, since you're not returning the expression (which is just assignment) you need to include curly braces

dane-johnson: its just going to return `undefined` anyhow

__tests__/actions/user.js
dane-johnson: You made up that word

solkaz: [😄](https://www.merriam-webster.com/dictionary/columnar)

thebho:
I'm tempted to merge this now before we get into merge conflict hell again.  Thoughts?

PR #419, opened 2017-04-19
Refactor Auth to dispatch thunks to authorize users with Github

### LINKED ISSUE #349: Handle Github authentication with Redux
## Expected Behavior
Authentication is handled with a redux action; credentials are stored in state (not necessarily eliminating the use of cookies)

## Current Behavior
Authentication is handled by the `Auth` component

### LINKED ISSUE #363: Migrate REST requests to be fired from Redux actions instead of from components
## Expected Behavior
REST requests are fired from redux actions, not from a component. 

## Current Behavior
REST requests are fired from components in `Auth` and `AppBody`

solkaz:
## Description
This PR refactors the `Auth` component to dispatch thunks to authorize a user with Github. Previously we would make AJAX requests directly on the component; instead the requests become abstracted away from the component, and simplifies the logic of logging in a user

This PR also refactors `AppBody` to similarly dispatch a thunk instead of directly making a request

## Motivation and Context
Fixes #349 
Fixes #363 

## Checklist:

### DIFF
--- a/__tests__/actions/user.js
+++ b/__tests__/actions/user.js
@@ -7,13 +7,13 @@ import * as actions from '../../src/actions/user';
 
 describe('Actions: User', () => {
   describe('action creators', () => {
-    it('creates correct SAVE_ACCESS_TOKEN object', () => {
+    it('creates correct SAVE_USERNAME object', () => {
       const token = 'token';
       const expected = {
-        type: actions.SAVE_ACCESS_TOKEN,
+        type: actions.SAVE_USERNAME,
         payload: token,
       };
-      expect(actions.saveAccessToken(token)).toEqual(expected);
+      expect(actions.saveUsername(token)).toEqual(expected);
     });
     it('creates correct SAVE_ACCESS_TOKEN object', () => {
       const token = 'token';
--- a/__tests__/actions/user.js
+++ b/__tests__/actions/user.js
@@ -6,6 +6,30 @@ import thunk from 'redux-thunk';
 import * as actions from '../../src/actions/user';
 
 describe('Actions: User', () => {
+  describe('action creators', () => {
+    it('creates correct SAVE_ACCESS_TOKEN object', () => {
+      const token = 'token';
+      const expected = {
+        type: actions.SAVE_ACCESS_TOKEN,
+        payload: token,
+      };
+      expect(actions.saveAccessToken(token)).toEqual(expected);
+    });
+    it('creates correct SAVE_ACCESS_TOKEN object', () => {
+      const token = 'token';
+      const expected = {
+        type: actions.SAVE_ACCESS_TOKEN,
+        payload: token,
+      };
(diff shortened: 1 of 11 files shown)


__tests__/actions/user.js
```diff
@@ -6,6 +6,30 @@ import thunk from 'redux-thunk';
...
+  describe('action creators', () => {
+    it('creates correct SAVE_ACCESS_TOKEN object', () => {
+      const token = 'token';
+      const expected = {
+        type: actions.SAVE_ACCESS_TOKEN,
+        payload: token,
+      };
+      expect(actions.saveAccessToken(token)).toEqual(expected);
+    });
+    it('creates correct SAVE_ACCESS_TOKEN object', () => {
```
Hopding: This appears to be an exact duplicate of the prior test. Would be good to remove it.

src/components/Auth.jsx
```diff
@@ -36,71 +41,43 @@ user is redirected to the URL they attempted to login from initially.
...
-    this.runLoginSequence = this.runLoginSequence.bind(this);
-    this.fetchAccessToken = this.fetchAccessToken.bind(this);
-    this.fetchUserInfo = this.fetchUserInfo.bind(this);
   }
 
   componentDidMount() {
     // Extract the code provided by GitHub from the redirect query string
     const { code } = this.props.router.location.query;
-    this.runLoginSequence(code);
+    // this.runLoginSequence(code);
```
Hopding: This line can be removed.

src/components/Auth.jsx
```diff
@@ -36,71 +41,43 @@ user is redirected to the URL they attempted to login from initially.
 export class Auth extends React.Component {
```
Hopding: Should `Auth` still be exported anymore, or should we only exported it after connecting it?

solkaz: We still use it for snapshot testing, which won't work with connected components

__tests__/components/Auth.test.jsx
```diff
@@ -8,9 +11,12 @@ import injectTapEventPlugin from 'react-tap-event-plugin';
 // http://stackoverflow.com/a/34015469/988941
 injectTapEventPlugin();
 
-import { Auth, errors } from '../../src/components/Auth';
+import { Auth, ConnectedAuth, errors } from '../../src/components/Auth';
```
Hopding: Does `Auth` need to be imported here?

solkaz: Yes, for snapshot testing

src/containers/CodesplainAppBar.jsx
```diff
@@ -58,9 +58,10 @@ export class CodesplainAppBar extends React.Component {
 
   handleSnippetSelected(key) {
     const username  = cookie.load('username');
+    console.log(username);
```
Hopding: Remove this log and the subsequent comments?

PR #425, opened 2017-04-20
Fix issue with snippet keys that contain apostrophes

### LINKED ISSUE #422: Apostrophes in snippet title cause URL routing to fail
## Expected Behavior
I should be able to add a snippet with a title `Brian's Python Example` and the backend should covert the title to a URL friendly key.  This was happening in previous releases, the title `Brian's Python Examples` is saved as `https://dev.codesplain.io/thebho/brian's_python_examples`.

## Current Behavior
In dev we now get  `https://dev.codesplain.io/thebho/brian%27s_python_copycat` when using the title: `Brian's Python Copycat` and that URL redirects back to the homepage.  There is a snippet saved with that key.

1. Create a new snippet with a title containing an apostrophe
2. Save that snippet
3. Reload the page

## Possible Solution
Trim the apostrophe before doing the url_lib.quote conversion in the lambda
Do a find and replace with a `-`
Add a UI workaround

solkaz:
## Description
This PR adds an enhanced encodeURIComponent function to URI encode characters that `encodeURIComponent` does not encode, such as apostrophes and parentheses. This function is used by `makeSaveEndpointUrl`, which is used by the saving and loading actions, meaning those work as intended now

## Motivation and Context
Fixes #422 

## Checklist:

### DIFF
--- a/__tests__/util/requests.test.js
+++ b/__tests__/util/requests.test.js
@@ -18,7 +18,7 @@ describe('util: requests', () => {
   });
   describe('sanitizeKey', () => {
     it('encodes characters that encodeURIComponent does not', () => {
-      const str = '!()*\'';
+      const str = './"!()*\'';
       const encoded = reqUtils.sanitizeKey(str);
       // Check that none of the characters in the original string appear in
       // the encoded string
--- a/__tests__/util/requests.test.js
+++ b/__tests__/util/requests.test.js
@@ -16,4 +16,18 @@ describe('util: requests', () => {
       expect(reqUtils.makeSaveEndpointUrl(username, snippet)).toEqual(expected);
     });
   });
+  describe('sanitizeKey', () => {
+    it('encodes characters that encodeURIComponent does not', () => {
+      const str = '!()*\'';
+      const encoded = reqUtils.sanitizeKey(str);
+      // Check that none of the characters in the original string appear in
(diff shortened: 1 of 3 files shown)


thebho:
Does this work with `'` `"` `.`  `/` ?

solkaz:
Yes

dane-johnson:
Doesn't this happen serverside?

solkaz:
Yes, but the issue is the key is decoded on the client side, so we need to encode it back to make the correct URI to load a snippet

dane-johnson:
Sounds convoluted

solkaz:
Apostrophes aren't URI-friendly, so if I have a key with an apostrophe (e.g. to load a snippet), the request will fail. So we need to encode apostrophes and other characters that `encodeURIComponent` does not so.

dane-johnson:
Here's my model
1. User enters title
2. Title is sent in the clear to server as JSON field
3. Server creates snippet key and returns to front end
4. Front end displays URI

Is this accurate?

thebho:
It does, but there are still things, like apostrophe, that get "fixed" by the server but the UI unfixes

thebho:
This works,`https://dev.codesplain.io/thebho/brian's_python_examples` and it was saved before we updated the key cleanser

PR #434, opened 2017-04-26
Snackbar

### LINKED ISSUE #333: Clean up the Saving Snippet UX
When saving a snippet the snackbar should display a saving icon or `...` animation.
On a successful save turn the snackbar `green`
On a failed save, turn the snackbar `red`

### LINKED ISSUE #348: Make a general-purpose notification system
## Expected Behavior
There should be one component to manage notifications for the user, either for logging in, saving, errors, etc. Messages can be sent to this component by dispatching redux actions, and they shouldn't be coupled to a container/component. For components that need interaction, such as when prompting for a title when doing 'Save As' don't necessarily have to use the same component/system, but it would be better if it did.

## Current Behavior
We have a Snackbar that is owned by the SnippetArea component, and has to be managed by it. Other components in the application cannot use it, and the SnippetArea now has to manage that component when it doesn't necessarily need to

solkaz:
# Description
This PR adds a general-purpose Snackbar component that isn't tied to any specific component. You can interact with it by dispatching an action to add a notification, and it will close that notification either when a user clicks outside of the snackbar, or the timeout duration (5 seconds) has passed

Also adds some refactoring/removal of dead code

## Motivation and Context
Fixes #348 
Fixes #333 

## Checklist:

### DIFF
--- a/__tests__/components/NotificationSnackbar.test.jsx
+++ b/__tests__/components/NotificationSnackbar.test.jsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { shallowToJson } from 'enzyme-to-json';
+import getMuiTheme from 'material-ui/styles/getMuiTheme';
+
+import { NotificationSnackbar, mapDispatchToProps } from '../../src/components/NotificationSnackbar';
+
+const defaultProps = {
+  displaying: false,
+  notification: 'Grasssssss tastes baddd',
+  onRequestClose: jest.fn(),
+};
+
+describe('<NotificationSnackbar />', () => {
+  const muiTheme = getMuiTheme();
+  const shallowWithContext = node => shallow(node, { context: { muiTheme } });
+
+  describe('snapshots', () => {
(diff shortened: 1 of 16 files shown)


__tests__/components/NotificationSnackbar.test.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+    describe('mapDispatchToProps', () => {
+      it('matches snapshot', () => {
+        expect(mapDispatchToProps).toMatchSnapshot();
+      });
+    });
+  });
+
+  describe('prop: displaying', () => {
+    it('is forwarded to the Snackbar\'s open prop', () => {
```
Hopding: Just curious, is there a reason you don't use ` `` ` strings to avoid the ugly `\'` escapes?

solkaz: Not really, I just prefer to use single quotes. I also don't find the `\'` escapes to be unreadable  enough to warrant using backticks

PR #456, opened 2017-05-04
Fix redirects to Home page from Not Found page

### LINKED ISSUE #450: Clicking on Codesplain from the `Resource Not Found` page doesn't redirect

Hopding:
## Description

* Convert `<NotFound />` to use `router.push` instead of `window.location`.
* Cause `<AppBody />` to run its snippet loading logic not only when first mounted, but also when the URL path changes.

## Motivation and Context

Fixes #450. Prior to this change, clicking on the header from the Not Found page would not take you back to the Home screen, despite the URL changing. This was due to using `router.push('/')` to handle redirects when the header was clicked, which resulted in `<AppBody />`'s state persisting across renders. This included the `isValidSnippet` state variable, which would have been set to false, and prevented the normal Home screen from displaying.

## Checklist:

### DIFF
--- a/src/containers/AppBody.jsx
+++ b/src/containers/AppBody.jsx
@@ -41,16 +41,9 @@ export class AppBody extends Component {
   }
 
   componentDidMount() {
-    const {
-      dispatch,
-      router,
-    } = this.props;
-    const {
-      id: snippetKey,
-      username,
-    } = router.params;
+    const { dispatch } = this.props;
 
-    // If the user is authenticated, add her Github to the orgs, and make it
+    // If the user is authenticated, add their Github to the orgs, and make it
     // the selected value
     if (cookie.load('token') && cookie.load('username')) {
       const savedUsername = cookie.load('username');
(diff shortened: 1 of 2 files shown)


src/containers/AppBody.jsx
```diff
@@ -37,18 +37,13 @@ export class AppBody extends Component {
...
-      dispatch,
-      router,
-    } = this.props;
-    const {
-      id: snippetKey,
-      username,
-    } = router.params;
+    const { dispatch } = this.props;
 
-    // If the user is authenticated, add her Github to the orgs, and make it
```
dane-johnson: It's singular, you should use a singular noun

dane-johnson: English's lack of neuter pronouns forces a gendered one.

src/containers/AppBody.jsx
```diff
@@ -59,6 +54,29 @@ export class AppBody extends Component {
...
+    const {
+      dispatch,
+      router,
+    } = this.props;
+    const {
+      id: snippetKey,
+      username,
+    } = router.params;
+
+    this.setState({ pathname: router.location.pathname });
```
solkaz: This shouldn't be necessary, per above

src/containers/AppBody.jsx
```diff
@@ -59,6 +54,29 @@ export class AppBody extends Component {
       cookie.load('orgs').split(' ').forEach(org => dispatch(addOrg(org)));
     }
 
+    this.loadSnippet();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    // Reload snippet if URL changes. Assume it is valid, and
+    // if it is not, the call to loadSnippet will handle it.
+    if (this.state.pathname !== nextProps.router.location.pathname) {
```
solkaz: Since you've wrapped the component using `withRouter`, you have access to the router from current props, so you can compare it with `nextProps` instead of having to save the pathname to state. So
```js
if (this.props.router.location.pathname !== nextProps.router.location.pathname) {
  // ...
}
```

Hopding: I tried that, however the router itself appears to mutate its state. E.g. If I go from `/Hopding/blah` to `/`, the value for `router.location.pathname` is `/` in both `this.props` and `nextProps`.

src/containers/AppBody.jsx
Hopding: See above response.

PR #462, opened 2017-05-08
Refactor application layout

### LINKED ISSUE #278: Prototype new layout of the application
As mention [here](https://github.com/maryvilledev/codesplainUI/pull/277#issuecomment-289517851)

I think the layout could be changed to make for a better UX

### LINKED ISSUE #414: The Select All and Clear All buttons need to take up less space when the window is smaller
The buttons need to be next to each other or moved over to the right edge of the Filters panel

## Current Behavior
screen shot 2017-04-18 at 7 56 09 am

### LINKED ISSUE #415: Long titles need padding between the language selector
## Current Behavior
screen shot 2017-04-18 at 8 16 14 am

### LINKED ISSUE #429: Fix contact area spillover
## Expected Behavior

The contact info should remain distinct from the other components.
## Current Behavior

Contact info is appearing atop the other components.
image_uploaded_from_ios_480

1. Resize the window

## Possible Solution

N/A

solkaz:
## Description
This PR refactors the layout of the application to be aesthetically pleasing

## Motivation and Context
Fixes #278 
Fixes #414
Fixes #415
Fixes #429

## Checklist:

screen shot 2017-05-08 at 8 50 12 am

screen shot 2017-05-08 at 8 41 18 am

screen shot 2017-05-08 at 8 41 37 am

screen shot 2017-05-08 at 8 42 09 am

screen shot 2017-05-08 at 8 42 40 am


thebho:
Photos?

thebho:
Is there a way to set a max rows of filters?  I worry a dozen filters would push the annotation panel down too far.  Maybe scrollable after 3 rows or do 3 wide annotation buttons.

solkaz:
Should be able to make the list scrollable. Might also consider resizing the filter buttons or panel, since they aren't nicely aligned

thebho:
What is the code mirror/right panel ratio right now.  What happens when we adjust it enough to do 3 wide filters at the current size?

Hopding:
Is the filters list opened or closed by default? 

Would it be possible to have the title, language selector, and buttons all in the same row? Reflow is still important, but two rows seems excessive at the width of the above screenshots.

src/components/SnippetAreaToolbar.jsx
```diff
@@ -61,33 +52,40 @@ const SnippetAreaToolbar = (props) => {
...
-          enabled={saveEnabled}
-          id="saveMenu"
-          onSaveAsClick={onSaveAsClick}
-          onSaveClick={onSaveClick}
-          style={styles.buttons}
-          orgs={orgs}
-          onOrgChanged={onOrgChanged}
-          selectedOrg={selectedOrg}
+      <div
+        style={{
```
Hopding: For consistency's sake, this style object should be moved into the `styles` constant above.

src/containers/App.jsx
```diff
@@ -5,7 +5,7 @@ import CodesplainAppBar from './CodesplainAppBar';
 import NotificationSnackbar from '../components/NotificationSnackbar';
 
 export const App = () => (
-  <div className="container-fluid">
+  <div style={{ height: 'inherit' }}>
```
Hopding: This style object should be moved into a `styles` constant.

src/containers/AppBody.jsx
```diff
@@ -137,31 +139,15 @@ export class AppBody extends Component {
       return <NotFound />;
     }
     return (
-      <div className="container-fluid">
-        <div className="row">
-          <div className="col-lg-2">
+      <div style={{ height: '85%' }}>
```
Hopding: See above comments.

src/containers/FilterArea.jsx
```diff
@@ -19,9 +19,22 @@ import CustomPropTypes from '../util/custom-prop-types';
 export class FilterArea extends React.Component {
   constructor(props) {
     super(props);
+    this.state = {
+      isExpanded: false,
```
Hopding: I think the default state of the filters should be expanded.

src/containers/FilterArea.jsx
```diff
@@ -19,9 +19,22 @@ import CustomPropTypes from '../util/custom-prop-types';
...
     this.handleRuleSelected = this.handleRuleSelected.bind(this);
     this.handleClearAllFilters = this.handleClearAllFilters.bind(this);
     this.handleSelectAllFilters = this.handleSelectAllFilters.bind(this);
+    this.handleExpandChange = this.handleExpandChange.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const numberOfFilters = Object.keys(this.props.filters).length;
+    const numberOfNextFilters = Object.keys(nextProps.filters).length;
+    // Set isExpanded to false if there are no filters to display
```
Hopding: Will it automatically expand if filters are added?

thebho:
Contact area needs some love
screen shot 2017-05-09 at 8 09 02 am

thebho:
Also looks like we have a regression with overlapping highlighting 
screen shot 2017-05-09 at 8 10 45 am

src/containers/FilterArea.jsx
dane-johnson: I concur

src/containers/FilterArea.jsx
dane-johnson: I think that would be annoying.

solkaz: I just changed it so that it will expand if there are filters added.

Hopding:
Looks good as soon as the highlighting overlap and contact area stuff is fixed 👍

thebho:
How does this look on mobile? It's not a big concern, but with a major redesign I would want to make sure we don't have any regressions there

thebho:
Also, since we are looking at the redesign and this is how we will probably release 1.0, I do think we should look into color.  Yes the Maryville color is that blue/teal, but there aren't a lot of solid panels with that color.  What if something like: https://maryvilledevcenter.com were `Codesplain` is that teal, and the head is black?

solkaz:
It works, other than the contact information section, which I can't figure out how to fix that issue

screen shot 2017-05-10 at 9 17 10 am
screen shot 2017-05-10 at 9 17 19 am

thebho:
Also, what does Codesplain look like *bold*?

solkaz:
I could look into changing the theme color, but that's probably more suited for another story, since this is concerned primarily with the layout of the application.

thebho:
Fair enough

solkaz:
As for a **bold** Codesplain:

screen shot 2017-05-10 at 9 20 43 am

Not bold:

screen shot 2017-05-10 at 9 21 29 am

thebho:
Hm, it's better, but maybe the font is wrong too.  Anyway, I added that to a new header story.  You are right, lets focus on layout here

PR #463, opened 2017-05-08
Don't show ignored rules in filters list

### LINKED ISSUE #461: Atoms are no longer being ignored in the UI
In older saved snippets, like: https://dev.codesplain.io/thebho/test3, atoms still show up under Filters, even tho they are no longer a filter option.

Hopding:
## Description

This PR prevents ignored filters from being displayed when a new set of mappings is being loaded. This will prevent 'Atoms' from being displayed as a filter (for now anyways - it is a persistent beast, and may show up again. Be wary...).

I also removed the call to [`window.location`](https://github.com/maryvilledev/codesplainUI/compare/IgnoreAtoms461?expand=1#diff-c872e8f6ec35e5dd45085d19b55c0892L89) when a snippet is loaded via a user's snippet list. This way we do not have to reload the whole app and make unnecessary AJAX calls, resulting in a faster snippet loading experience.

## Motivation and Context

Fixes #461. Prior to this change, when a snippet was loaded that required a new set of mappings to be loaded, the `removeDeprecatedFilters` function would fail to work properly, as the `ignoredRules` would not yet have been set.

## Checklist:


src/components/Editor.jsx
```diff
@@ -43,11 +43,7 @@ const makeMarker = () => {
   return marker;
 };
 
-const pushValueToCodeMirror = _.once((value, codeMirrorInst) => {
-  if (!codeMirrorInst.getValue() && value) {
-    codeMirrorInst.setValue(value);
-  }
-});
+const pushValueToCodeMirror = (value, codeMirrorInst) => codeMirrorInst.setValue(value);
```
dane-johnson: Why?

Hopding: > I also removed the call to window.location when a snippet is loaded via a user's snippet list. This way we do not have to reload the whole app and make unnecessary AJAX calls, resulting in a faster snippet loading experience.

This means that the value may need to be pushed more than once in a session, because multiple snippets can now be loaded in a single session.

dane-johnson: But why remove the guard?

Hopding: Because the CodeMirror will have a value, but it will be that of the older snippet.

thebho: @dane-johnson Do you still have concerns about this?

dane-johnson: If I works I don't.

PR #464, opened 2017-05-09
Implement snippet selector lists for organizations

solkaz:
## Description
This PR implements the ability to select an organization's snippet from the `AppMenu` component akin to selecting your own snippets.

## Motivation and Context
Fixes #445 

## Checklist:

- [x] My code follows the code style of this project.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

## Screenshots (if appropriate):
screen shot 2017-05-09 at 12 51 49 pm
screen shot 2017-05-09 at 12 51 56 pm
screen shot 2017-05-09 at 12 52 08 pm

PR #473, opened 2017-05-11
Move user data from cookies to redux

### LINKED ISSUE #447: Logged in users can be locked out of features when the Auth/Org list logic is updated
This has happened to me multiple times, with the most recent being the worst.  I didn't save the console log, but essentially I was already logged in and couldn't load any snippets because the the org list was `undefined`  Logging out and back in solved the problem

We could solve this by forcing a re-log in after a new release

Hopding:
## Description

This PR moves the storage location of certain user data (such as their username, avatarURL, & orgs) from  cookies to the redux state. This resolves the problems we have had with stale data, as it causes the data to be updated every time the UI is loaded. The only data that is being stored in a cookie, now, is the user's `token` (and the `signInState` & `signInRedirect`). The `<CodesplainAppBar />` component is in charge of loading the `token` cookie and updating the redux state with the user's info for use by the other components.

## Motivation and Context

Fixes #447. Prior to this change, we already held some of this information in redux, but not all of it, and we still kept the data in the cookies as well. This resulted in problems with stale data, as certain information (e.g. the user's organizations) was only updated when they logged in.

## Checklist:


dane-johnson:
This PR has a lot in common with #464. Is there functionality added here that is not there?

src/containers/CodesplainAppBar.jsx
```diff
@@ -50,8 +59,33 @@ export class CodesplainAppBar extends Component {
...
+          const { login: username, avatar_url: userAvatarURL } = res.data;
+          dispatch(saveUsername(username));
+          dispatch(setAvatarUrl(userAvatarURL));
+
+          // Add user's username to orgs list, and select it as default
+          dispatch(addOrg(username));
+          dispatch(switchOrg(username));
+        })
+        .then(() => dispatch(updateUserSnippets()));
+    }
```
solkaz: It would be beneficial to have a catch clause in case the token isn't valid. Specifically, it should remove the token from cookies and redirect to the home page

src/containers/CodesplainAppBar.jsx
```diff
@@ -185,23 +217,34 @@ export class CodesplainAppBar extends Component {
...
   hasUnsavedChanges: PropTypes.bool.isRequired,
   userSnippets: CustomPropTypes.snippets,
+  username: PropTypes.string.isRequired,
+  token: PropTypes.string,
+  avatarURL: PropTypes.string,
 };
 
 CodesplainAppBar.defaultProps = {
   userSnippets: {},
+  token: undefined,
```
solkaz: This should probably be an empty string, since it is a string type  (and will evaluate to false, so it won't break any of your code)

Hopding:
@dane-johnson It does. #464 does not remove the app's usage of cookies for storing user data. This PR goes. Look specifically at the changes made to `components/Auth.jsx` and `containers/CodesplainAppBar.jsx`.

PR #495, opened 2017-05-15
Snippet owner icon is removed when the user navigates away from a saved snippet

### LINKED ISSUE #485: When a user navigates away from a saved snippet, the snippet owner icon should be removed
## Expected Behavior
Leaving the context of a saved snippet, should completely reset the state of the app

## Current Behavior
screen shot 2017-05-12 at 8 15 45 am

1. Go to a saved snippet
2. Click the `Codesplain` title to navigate back to home

solkaz:
## Description
- Refactors the snippet author avatar into its own component `AuthorAvatarIcon`.
- Reset the snippet author when navigating away from a snippet
- When loading another snippet, if that snippet's author is the logged-in user, use the avatar URL in the Redux store instead of pinging Github for it.

## Motivation and Context
Fixes #485

## Checklist:


src/components/AuthorAvatarIcon.jsx
```diff
@@ -0,0 +1,52 @@
...
+    paddingTop: '10px',
+    verticalAlign: 'bottom',
+  },
+};
+
+const AuthorAvatarIcon = ({ author, avatarUrl }) => {
+  if (avatarUrl) {
+    return (
+      <div
+        data-for="avatar"
```
Hopding: I think it might be better to change this to `data-for="author-avatar"`. If we ever added a tooltip to an `<Avatar />` component, we would likely use `data-for="avatar"`, which would clash with this div's current `data-for` value.

PR #498, opened 2017-05-15
Optimize highlight.js imports

solkaz:
## Description
This PR refactors our imports from `highlight.js`, eschewing the default instance (which is around 520 kb when imported) and instead using a hand-rolled solution that supports only the Java and Python languages and is 12 kb. 

## Motivation and Context
To decrease our bundle sizes, the benefits of which should be obvious.

## Checklist:

## Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Mobile:
- [ ] Other:

Before:
`Highlight.js` bundled size: (520kb)
screen shot 2017-05-15 at 3 44 24 pm

Total bundle size: (~1.5mb)
screen shot 2017-05-15 at 3 44 34 pm

After:
`Highlight.js` bundled size: (~12kb)
screen shot 2017-05-15 at 3 44 44 pm

Total bundle size: (~1mb)
screen shot 2017-05-15 at 3 44 54 pm

### DIFF
--- a/src/util/highlight.js
+++ b/src/util/highlight.js
@@ -1,0 +1,13 @@
+// Default hljs isn't imported as it contains ALL languages (which totals to
+// ~520 kb).
+import hljs from 'highlight.js/lib/highlight';
+
+// Each language needs to be imported like so:
+import java from 'highlight.js/lib/languages/java';
+import python from 'highlight.js/lib/languages/python';
+
+// Register each language to the hljs instance
+hljs.registerLanguage('java', java);
+hljs.registerLanguage('python', python);
+
+export default hljs;
--- a/src/util/highlight.js
+++ b/src/util/highlight.js
@@ -3,10 +3,12 @@
 import hljs from 'highlight.js/lib/highlight';
 
 // Each language needs to be imported like so:
+import go from 'highlight.js/lib/languages/go';
 import java from 'highlight.js/lib/languages/java';
 import python from 'highlight.js/lib/languages/python';
 
 // Register each language to the hljs instance
+hljs.registerLanguage('go', go);
 hljs.registerLanguage('java', java);
 hljs.registerLanguage('python', python);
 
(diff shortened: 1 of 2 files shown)


Hopding:
I think this might remove too many languages. I think we should also support more of the common languages, specifically:

* `swift`
* `javascript`
* `cpp`
* `objectivec`
* `haskell`
* `bash`
* `xml`
* `ruby`

solkaz:
Open to adding new languages, but we don't have users that are requesting those languages. Plus, it's easy to add a language; this was done to cut out the bloat.

thebho:
Did you really try this on all the browsers ?

dane-johnson:
@Hopding is right. People may want to write annotations with inlined code from a language we don't support, perhaps as a point of comparison. If bundle size is really a problem, we should find another way to reduce it.

solkaz:
@thebho just on Chrome and Safari

thebho:
Then why did you check them all?

solkaz:
If we want to add more languages, then sure, we can do that later, and it's an easy process that's evident in the code. Right now, however, we don't have any users for Codesplain. We ourselves don't, and probably won't, use the annotation feature for other languages. That's why I don't feel a need to include more languages; if you really, _really_ want a language, you're welcome to make a PR.

solkaz:
@thebho because I imagined that this should work across all browsers. It's not a UI change.

thebho:
Just add the browsers you have actually tested the changes on. I'm not expecting every PR to have full coverage, but when there are layout or state changes, we should have a record of the browsers that have been tested

dane-johnson:
Would like to see support for most of the top 30
http://spectrum.ieee.org/computing/software/top-10-programming-languages/

solkaz:
@dane-johnson 
> if you really, _really_ want a language, you're welcome to make a PR.

PR #501, opened 2017-05-16
Remove Bootstrap dependency

### LINKED ISSUE #491: Remove Bootstrap
Bootstrap is still a dependency and is pulled in when the user loads the page, adding to the load times. However, we don't use any of Bootstrap's CSS classes, so there's no reason we need to pull it in. This isn't as simple as removing the tags, because Bootstrap affects how certain tags are styled and thus there are noticeable differences once we remove it.

solkaz:
## Description
- Remove Bootstrap and its associated dependencies from the app
- Add style changes to accommodate the styling lost with removing bootstrap.

## Motivation and Context
Fixes #491 

## Checklist:

## Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Mobile:
- [ ] Other:

Before (Chrome):
screen shot 2017-05-16 at 11 06 32 am

After:
screen shot 2017-05-16 at 11 06 18 am

### DIFF
--- a/__tests__/components/__snapshots__/AnnotationEditor.test.jsx.snap
+++ b/__tests__/components/__snapshots__/AnnotationEditor.test.jsx.snap
@@ -37,6 +37,7 @@ exports[`<AnnotationEditor /> matches snapshot 1`] = `
         "alignItems": "center",
         "display": "flex",
         "flexFlow": "row wrap",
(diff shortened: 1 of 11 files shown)


thebho:
Does this change anything with smaller window sizes?

PR #503, opened 2017-05-16
Optimize lodash imports

solkaz:
## Description
Optimizes lodash imports by selectively importing only the functions we need (as opposed to importing the default lodash object and calling methods from it).

## Motivation and Context
Cuts the bundled lodash code from 71.3kb to 20.2kb in the Web Worker and 31.2kb. Whereas lodash was 71.5% of the Web Worker bundle, it is now 41.6%, meaning it will load much faster (important for highlighting and fetching parsers)

## Checklist:

## Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Mobile:
- [ ] Other:


src/components/Editor.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react';
 import CodeMirror from 'react-codemirror';
-import _ from 'lodash';
+import isEqual from 'lodash/isEqual';
```
dane-johnson: I read somewhere this might not be the best idea. Standby for source...

solkaz: [Not so](http://stackoverflow.com/a/35251059)

solkaz: Our current method, which is
```js
import _ from 'lodash';
```
will import _all_ of lodash (which is about 75kb), including the functions we don't use. This method instead imports only those functions we want.

dane-johnson: Yeah I get that but you're choosing to spare the memory and sacrifice speed this way

solkaz: We're not sacrificing speed because it performs exactly the same as it did before. The only difference this PR introduces is the decreased bundle size.

dane-johnson: More imports means more load time. I think in the worker where it only uses a few that is fine but in proper that's a lot of imports that can be consolidated.

PR #508, opened 2017-05-17
Fix errors and warnings in console

### LINKED ISSUE #507: The snippet list isn't being updated when deleting a snippet
## Expected Behavior
After deleting a snippet, the snippet lists should be refreshed in the browser

## Current Behavior
After deleting a snippet, that snippet is still present in the My Snippets menu, and doesn't go away until you refresh the browser

## Browser
Chrome

1. Open a snippet
2. Delete that snippet
3. Open the `My Snippets` menu
4. See that the snippet name is still on the list

Hopding:
## Description

* Fixed import location of `setAuthor` in `SnippetArea`, so the user is properly redirected and their snippet lists are updated when they delete a snippet.
* Fixed the URL being used to get a user's avatar, which was resulting in a 404.
* Fixed PropType warning due to the app being rendered before `snippetKey` was set.

## Motivation and Context

Fixes #507. Nobody likes type errors, 404s, and PropType warnings.

## Checklist:

## Browsers

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Mobile:
- [ ] Other:

### DIFF
--- a/src/util/requests.js
+++ b/src/util/requests.js
@@ -43,15 +43,15 @@ export const sanitizeSnippetList = (snippetList) => {
   return snippetList;
 };
 
-export const makeUserUrl = user => `${GITHUB_API_URL}/users/${user}`;
+export const makeUserUrl = () => `${GITHUB_API_URL}/user`;
 
-export const fetchUserAvatar = (user, token) => {
+export const fetchUserAvatar = (token) => {
   const reqHeaders = {
     headers: {
       Authorization: `token ${token}`,
     },
   };
-  const reqUrl = makeUserUrl(user);
+  const reqUrl = makeUserUrl();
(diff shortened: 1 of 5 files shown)


src/util/requests.js
```diff
@@ -43,15 +43,15 @@ export const sanitizeSnippetList = (snippetList) => {
   return snippetList;
 };
 
-export const makeUserUrl = user => `${GITHUB_API_URL}/users/${user}`;
+export const makeUserUrl = () => `${GITHUB_API_URL}/user`;
```
dane-johnson: You can't do this because in this context `user` isn't necessarily the authenticated user, can also be the owner of the snippet.

Hopding:
@dane-johnson That problem is fixed now as a result of pulling in master & resolving the conflicts.

PR #519, opened 2017-05-19
Logout users with invalid tokens

### LINKED ISSUE #465: Logout users with expired tokens
## Expected Behavior

Users should be asked to log back in if they have expired/invalid tokens.

## Current Behavior

Currently, if a user performs an operation that results in a call to the API, and the API responds with a `400` due to an invalid auth token, the user is not notified of _why_ the operation failed, but only the fact of it. Instead, they should be prompted to log back into the app in order to obtain a valid auth token.

## Possible Solution

Create a redux action that when invoked, logs the user out of the app, and displays a modal requesting that they log back in. Parse all responses from API requests that can fail due to invalid tokens, and dispatch that action when they fail because of an invalid token.

solkaz:
## Description
- When a user attempts to access their information with an invalid Github token, the application will clear their credentials and reset the application. This will occur early in the application life cycle, specifically in `<CodesplainAppBar />`'s `componentDidMount` method.
- Abstracts much of the promise logic away from the component, and instead lets the thunks handle them.

- This PR relies on [this PR](https://github.com/maryvilledev/codesplain-lambdas/pull/117) being merged in order to work correctly

## Motivation and Context
Fix #465 

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] All new and existing tests passed.

## Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari

### DIFF
--- a/__tests__/containers/CodesplainAppBar.test.jsx
+++ b/__tests__/containers/CodesplainAppBar.test.jsx
@@ -6,6 +6,7 @@ import { shallowToJson } from 'enzyme-to-json';
 import getMuiTheme from 'material-ui/styles/getMuiTheme';
 import injectTapEventPlugin from 'react-tap-event-plugin';
 
+import { saveAccessToken } from '../../src/actions/user';
 import generateMockStore from '../../src/testUtils/mockStore';
 import generateMockRouter from '../../src/testUtils/mockRouter';
 import {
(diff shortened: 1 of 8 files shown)


thebho:
Why did you refactor all the gists stuff?

solkaz:
@thebho part of refactoring the CodesplainAppBar, so that most of the logic is abstracted away.

solkaz:
This "works", but I cannot figure out a way to test this. Merge at your own discretion

thebho:
Do you mean you don't have a way to write unit tests, or you have never actually seen the code work?

solkaz:
I don't have a way to write unit tests. I can verify that this works in all browsers, though

PR #525, opened 2017-05-22
Fix responsiveness of the RuleLabel buttons

### LINKED ISSUE #517: Windows need to be 1400w before the Filters are displayed in 2 columns
## Expected Behavior
At 1200 the filters should still be displayed in 2 columns

## Current Behavior
It varies, but usually the filters display in a single column at less than 1400 width

## Browser
Safari and Chrome

solkaz:
## Description
Refactors the RuleLabel buttons to use flex-box properties, making them more responsive
Also removes the padding from the Filter and Annotation titles, so there's more room for the content in their respective sections

## Motivation and Context
Fixes #517

## Checklist:

## Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Mobile:
- [ ] Other:

### DIFF
--- a/src/components/RulesSelector.jsx
+++ b/src/components/RulesSelector.jsx
@@ -29,7 +29,7 @@ const makeListItems = (filters, onRuleSelected) => (
 const RulesSelector = ({ filters, onRuleSelected }) => {
   const listItems = makeListItems(filters, onRuleSelected);
   return (
-    <div>
+    <div className="rule-selector-container">
       {listItems}
     </div>
   );
--- a/src/components/RulesSelector.jsx
+++ b/src/components/RulesSelector.jsx
@@ -1,5 +1,7 @@
 import React, { PropTypes } from 'react';
+
 import RuleLabel from './RuleLabel';
+import CustomPropTypes from '../util/custom-prop-types';
 
 const makeListItems = (filters, onRuleSelected) => (
   Object.keys(filters)
@@ -34,12 +36,7 @@ const RulesSelector = ({ filters, onRuleSelected }) => {
 };
 
 RulesSelector.propTypes = {
-  filters: PropTypes.shape({
-    color: PropTypes.string,
-    count: PropTypes.number,
-    prettyTokenName: PropTypes.string,
-    selected: PropTypes.bool,
-  }).isRequired,
+  filters: CustomPropTypes.filters.isRequired,
   onRuleSelected: PropTypes.func.isRequired,
 };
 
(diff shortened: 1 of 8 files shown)


src/components/RulesSelector.jsx
```diff
@@ -27,19 +29,14 @@ const makeListItems = (filters, onRuleSelected) => (
 const RulesSelector = ({ filters, onRuleSelected }) => {
   const listItems = makeListItems(filters, onRuleSelected);
   return (
-    <div>
+    <div className="rule-selector-container">
```
dane-johnson: We are doing styles as an object, not css files

solkaz: We're using both of those methods in the project, this isn't new. Also, it's more conducive for development since you don't have to reload the app to see your changes.

dane-johnson: @Hopding is this true? I've been doing all of my styling as objects

solkaz: You can check in `AppBody`; both the SnippetArea and the right section use classes

Hopding: My understanding is that we decided to use objects for styling. I'm not aware of any files that use CSS classes (did a quick search and didn't find any either), with the exception of `Editor.jsx`. However, the reason that component uses them is because it is required for CodeMirror styling.

Hopding:
This is definitely a nice improvement. However you removed the padding between the filters. Was that intentional? I'd prefer to keep it.

thebho:
It was a compromise to use dynamic sizing

src/components/RulesSelector.jsx
solkaz: I meant to say that the elements in AppBody are using CSS id selectors, not classes

In this case, we're not required to use CSS selectors to apply styles to the buttons, although I think it's more appropriate to do so in this case for a couple of reasons. That said, I can revert to using our previous method.

Hopding:
Why is padding incompatible with resizing?

src/components/RulesSelector.jsx
Hopding: There are certain CSS selectors that don't directly translate to JavaScript object. So if its going to be complicated to use objects for certain selectors, then I'm not against keeping them. But for the sake of consistency, it would be good to try to use objects when possible.

solkaz:
*Margin, not padding. For some reason, adding margins breaks flexbox properties.

thebho:
Yeah, this was the best of the options i saw.  There is a slack history from yesterday if you are interested

PR #531, opened 2017-05-23
Hotfix to ignore rules that existed in legacy parsers but not now

dane-johnson:
## Description

More effective guard against non-understood rules. Very temporary solution, these rules should never be able to exist in the first place

## Motivation and Context

Closes maryvilledev/codesplain#97

## Checklist:

## Browsers

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Mobile:
- [ ] Other:


1jordan:
Please test in other browsers.

PR #534, opened 2017-05-24
Codesplain will now balk when asked to load the same snippet

### LINKED ISSUE #533: When selecting the current snippet from the snippet menu, the state is erased
## Expected Behavior
The state should refresh to match the saved snippet

## Current Behavior
The state is wiped but the URL is correct.
Navigating to a different snippet doesn't cause this issue

## Browser
Chrome

1. Save a snippet or open a saved snippet
2. Go to `Snippets` -> `My Snippets` 😢  -> `CURRENT_SNIPPET_NAME`
3. See that the application state is gone

dane-johnson:
## Description

When asked to load a snippet that is already loaded, Codesplain will now skip reloading the snippet.

## Motivation and Context

Closes #533 

## Checklist:

## Browsers

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Mobile:
- [ ] Other:

### DIFF
--- a/src/containers/CodesplainAppBar.jsx
+++ b/src/containers/CodesplainAppBar.jsx
@@ -167,9 +167,11 @@ export class CodesplainAppBar extends Component {
   }
 
   handleSnippetSelected(snippetOwner, snippetKey) {
-    const { router } = this.props;
-    this.resetApplication();
-    router.push(`/${snippetOwner}/${snippetKey}`);
+    const { router, snippetKey: curr } = this.props;
+    if (snippetKey !== curr) {
+      this.resetApplication();
+      router.push(`/${snippetOwner}/${snippetKey}`);
+    }
   }
 
   handleTitleTouchTap() {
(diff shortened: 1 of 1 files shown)


thebho:
Will this just ignore the request, or actually reload the page with the state saved in S3?

dane-johnson:
It will ignore the request.

thebho:
That feels wrong to me.  If we don't want them selecting the current snippet, it should be disabled not ignored.  If you select a snippet it should force a reload (and a confirmation modal if anything has changed)

Hopding:
Being disabled or ignored is fine with me. I don't think we should refresh though. We do not refresh the page when you select a snippet from the list. We just perform an AJAX request and display the resulting data, but no browser refresh occurs. I see no reason that we should perform a browser refresh if the snippet you click happens to be the same one you're already viewing.

thebho:
Ok, no browser refresh, but it should still recover the state to the saved version if you make any changes.

PR #540, opened 2017-05-24
Add check when loading new parser for differing filters

solkaz:
## Description
When loading a new parser, checks if the new filters returned from `removeDeprecatedFiltersFromState` differs from the old filters; if it hasn't changed, then `restoreState` will not be dispatched

## Motivation and Context
Minor optimization

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] All new and existing tests passed.

## Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Mobile:
- [ ] Other:

### DIFF
--- a/src/util/rules.js
+++ b/src/util/rules.js
@@ -1,4 +1,5 @@
 import axios from 'axios';
+import isEqual from 'lodash/isEqual';
 import last from 'lodash/last';
 import {
   setRules,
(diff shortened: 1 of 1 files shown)


thebho:
What exactly does this fix?

solkaz:
Currently, when a new parser is loaded, `removeDeprecatedFiltersFromState` will be called on the current app state, and then `restoreState` will be dispatched after that. This is unnecessary if the filters object hasn't changed, though, so this adds a check to see if it's necessary to call `restoreState`, which can prevent an unnecessary re-render

solkaz:
It's just an optimization, it doesn't "fix" any bugs

PR #542, opened 2017-05-24
Snippet search

### LINKED ISSUE #489: Add a search field to filter snippets by title
This search field would sit at the top of a list of a user's snippets and the lists of org snippets

dane-johnson:
## Description

You can use fuzzy searching to find a snippet across all the snippets that you have access to.

## Motivation and Context

Closes #489 

## Checklist:

## Browsers

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Mobile:
- [ ] Other:

screen shot 2017-05-24 at 4 16 29 pm

### DIFF
--- a/src/containers/CodesplainAppBar.jsx
+++ b/src/containers/CodesplainAppBar.jsx
@@ -79,13 +79,13 @@ const orderSnippets = (username, userSnippets, orgSnippets) => {
   const allSnippets = { ...orgSnippets }; // Spread so there aren't side effects.
   allSnippets[username] = userSnippets;
   const orderedSnippets = _.reduce(allSnippets, (allList, snippets, role) => {
-    const next = _.reduce(snippets, (snipList, { language }, title) => {
-      const obj = { role, title, language };
+    const next = _.reduce(snippets, (snipList, { language, lastEdited }, snippetTitle) => {
+      const obj = { role, snippetTitle, language, lastEdited };
       return _.concat(snipList, obj);
     }, []);
     return _.concat(allList, next);
   }, []);
-  return _.sortBy(orderedSnippets, ({ title }) => title);
+  return _.sortBy(orderedSnippets, ({ snippetTitle }) => snippetTitle);
 };
 
 /*
--- a/src/containers/CodesplainAppBar.jsx
+++ b/src/containers/CodesplainAppBar.jsx
@@ -258,7 +258,14 @@ export class CodesplainAppBar extends Component {
       />,
     ];
 
-    const { avatarUrl, orgSnippets, username, userSnippets, gists } = this.props;
+    const {
+      avatarUrl,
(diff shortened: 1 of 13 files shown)


src/containers/CodesplainAppBar.jsx
```diff
@@ -73,6 +75,19 @@ const styles = {
...
+  allSnippets[username] = userSnippets;
+  const orderedSnippets = _.reduce(allSnippets, (allList, snippets, role) => {
+    const next = _.reduce(snippets, (snipList, { language, lastEdited }, snippetTitle) => {
+      const obj = { role, snippetTitle, language, lastEdited };
+      return _.concat(snipList, obj);
+    }, []);
+    return _.concat(allList, next);
+  }, []);
+  return _.sortBy(orderedSnippets, ({ snippetTitle }) => snippetTitle);
+};
```
solkaz: Can you clarify this code block?

thebho:
I like the functionality, but something feels unfinished with the modal.  Maybe it could use a black header and an `x` like the Centred popup (Lightbox) on this sight https://wet-boew.github.io/v4.0-ci/demos/overlay/overlay-en.html

src/containers/CodesplainAppBar.jsx
```diff
@@ -7,6 +7,7 @@ import {
 } from 'material-ui';
 import { withRouter } from 'react-router';
 import cookie from 'react-cookie';
+import _ from 'lodash';
```
dane-johnson: I definitely need it for functional programming.

src/containers/CodesplainAppBar.jsx
dane-johnson: I'll add a comment, it builds a vector out of the nested objects that contains only the details relevant to the search menu, and is sortable by name.

src/containers/CodesplainAppBar.jsx
solkaz: I'm not saying you cannot use lodash, but rather that importing the default lodash object isn't allowed. Import only the functions you need.

src/containers/CodesplainAppBar.jsx
```diff
@@ -7,6 +7,7 @@ import {
 } from 'material-ui';
 import { withRouter } from 'react-router';
 import cookie from 'react-cookie';
+import { reduce, concat, sortBy } from 'lodash';
```
solkaz: Check the other files, like [here](https://github.com/maryvilledev/codesplainUI/blob/master/src/util/annotations.js#L1) as an example of how you should import lodash. Even though you're only importing a few functions from the default lodash object, you still have to import the entire thing, so it effectively is the same as before. Instead, it should be:
```js
import reduce from 'lodash/reduce';
/// ...
```

